PR #3, opened 2015-04-03
claiming a PID

aidv:
am I doing it right?

### DIFF
--- a/1209/1033/index.md
+++ b/1209/1033/index.md
@@ -1,0 +1,9 @@
+---
+layout: pid
+title: AnyThing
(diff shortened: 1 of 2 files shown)


Arachnid:
Sorry, but you need to have a project with published source, and a device selected, before you can claim a PID. You can't claim one prospectively. Don't worry, they won't run out.

Arachnid:
Apologies, I shouldn't have accepted this: all PIDs starting with 0x1xxx are reserved. Please modify your pull request to use a different PID, and I'll be happy to accept it.

PR #6, opened 2015-04-03
added telavivmakers.org to the list

5shekel:
thank you for your generosity!


Arachnid:
Apologies, I shouldn't have accepted this: all PIDs starting with 0x1xxx are reserved. Please modify your pull request to use a different PID, and I'll be happy to accept it.

PR #9, opened 2015-04-03
request pid 2222 & 3333 for LabConnect

do9xe:
Malte and I would like to request the PIDs 1337 and 1338 for our Final Project we managed to get OpenSource before fininshing it.
Regards, the LabConnect Team


Arachnid:
We'd be glad to issue you PIDs for these projects. However, PIDs starting 0x1XXX are reserved by the original owners of VID 0x1209. Please select another pair of PIDs outside that range.

do9xe:
Well, as i see you didn't write anything about it any where ;) Ill correct it :) wat is the area for PIDs? :)

do9xe:
Ill commit a new PID right away

Arachnid:
Per `http://pid.codes/howto/`, "You may request any number that is not in use, except those reserved as described on the VID page.". I've since updated the text to be clearer that the range starting with 0x1xxx is reserved.

do9xe:
Update :) Hope I'll be fine :)

Arachnid:
Thanks for your patience while we work out the kinks in the process.

do9xe:
We thank you for the PID!!! :) Have a great time :)

PR #19, opened 2015-04-04
Hopefully getting a USB VID & PID for Omega Brewing! cross your fingers ...

markwbrown:
...for 2112!

### DIFF
--- a/1209/2112/index.md
+++ b/1209/2112/index.md
@@ -1,0 +1,8 @@
+---
+layout: pid
+title: Brewer
+owner: omega
+license: LGPL-3.0
+site: http://www.omegabrewing.com
+source: http://github.com/markwbrown/rasardbru/
+---
--- a/org/omega/index.md
+++ b/org/omega/index.md
@@ -1,0 +1,5 @@
+---
+layout: org
+title: Omega Brewing
+---
+We're a small brewery and hardware manufacturer


Arachnid:
Thanks for your pull request! However, we have identified a couple of issues preventing us from issuing you a PID:
 - Your project doesn't appear to use USB, and thus doesn't require a PID
 - Your repository does not contain working source or schematics for a USB device
 - You haven't added a LICENSE file to your repository

Please feel free to reopen this PR or open a new one if these issues are resolved.

markwbrown:
I think the code I have on Github is horribly outdated for my own repos, but we have a project we really would like to have a PID for, though isn't ready to go live on the internet, as it some of it relies on some code we have not been able to license yet.

As far as working source or schematics for a USB device, we're not there yet. There is still some more work to be done in Eagle and with PCB prototypes with respect to this. I hope that does not prevent us from qualifying for a PID

Where in the repo should the license file be, and what are you expecting? (e.g. LICENSE, license, license.txt, etc)

Arachnid:
The license file can be anywhere. Github standardises on LICENSE, but as long as it's clear to users what the license is, that's fine.

If you're not yet ready to release source or schematics, you're welcome to use our test VID/PID pair, 0x1209/0x0001 for internal testing; as long as you publish source and schematics before you start distributing hardware to users, we can issue you a PID then.

PR #24, opened 2015-04-04
Unethi

con-f-use:
PID request for unethi


Arachnid:
Thanks for your pull request!

You need to specify 'layout: pid', not 'layout: C0F5'. Otherwise, this looks good.

con-f-use:
Done. Thanks for the quick reply.

PR #29, opened 2015-04-04
added pid for ww

willemwouters:


Arachnid:
Thanks for your pull request!

There are a couple of issues that need to be resolved before we can issue you a PID:
 - Your project doesn't appear to use USB. Can you please clarify what you need the PID for?
 - You need to check in a LICENSE file in your source code repository to make it clear to users what license your project is under.
 - Please read carefully the directions at http://pid.codes/howto/; specifically, you need to check in an organisation file with the correct name.

PR #30, opened 2015-04-04
Add no-route.org and Fab Lab Berlin.

DatanoiseTV:


Arachnid:
I assume you intend to request PIDs for these organisations at a later stage? As it is, they won't show up anywhere.

Also, can you please delete the apple kruft from your PR? :)

DatanoiseTV:
Yes, that is correct. A lot of projects are ongoing in the fablab and by no-route.org and we would love a little PID pool if it's possible. I also deleted the .DS_Store trash.

Arachnid:
Generally pid.codes' policy is to issue PIDs only to concrete projects with existing source code. If you're happy with that, I'm happy to merge this for use by future PID applications.

Or, if you think you've got a compelling argument for pre-issuing a group of PIDs, let me know. :)

Arachnid:
Would you like this committed as-is?

PR #34, opened 2015-04-04
Requesting my PID for future use.

andernoo:
My designs are still rudimentary but they will need a PID in the future, so I thought I should stake my claim on one. Thanks for doing this for the maker community!


Arachnid:
I'm sorry, but you need to have a project with published schematics or source, using USB, and licensed appropriately. Please submit a new PR when your project meets these requirements; don't worry, there'll still be plenty of PIDs left.

PR #36, opened 2015-04-04
Add Mooltipass offline password manager

diafygi:


Arachnid:
Do you have any involvement with / authority over the mooltipass project?

diafygi:
I do not. I just wanted to reserve a slot for Mooltipass in case @limpkin wanted to get a pid before they all get used up.

PR #40, opened 2015-04-04
Keyboardio Model 01 & Keyboardio Model 01 Bootloader

obra:
Requesting a pair of PIDs for the Keyboardio Model 01 hackable ergonomic keyboard. As is customary, we'd like to use separate PIDs for the bootloader and user runtime. (They're separate codebases with separate USB profiles. They do run on the same hardware.)

Thanks so much!

### DIFF
--- a/1209/2300/index.md
+++ b/1209/2300/index.md
@@ -1,0 +1,8 @@
+---
+layout: pid
+title: Keyboardio Model 01 Bootloader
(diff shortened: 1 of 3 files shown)


Arachnid:
This looks like a fantastic project. Good luck!

obra:
Thanks!

PR #42, opened 2015-04-04
Add smcleod PID 1337

sammcj:


Arachnid:
Hi,

Thanks for your pull request! However, there are several issues preventing us from issuing you a PID:
 - Your source link is 404; there's nothing there!
 - PIDs in the range 0x1000 - 0x1FFF are reserved, as documented here: http://pid.codes/1209/.

Please feel free to open a new pull request at such point as these issues are resolved.

PR #44, opened 2015-04-05
Added information for my XFX RMC-1 midi controller.

AdamFreemer:
Added information for my XFX RMC-1 midi controller.

### DIFF
--- a/1209/2001/index.md
+++ b/1209/2001/index.md
@@ -1,0 +1,8 @@
+---
+layout: pid
+title: RMC-1
+owner: /org/xfx
+license: MIT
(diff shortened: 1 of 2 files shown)


Arachnid:
Thanks for your pull request! However, there are a couple of issues preventing us from issuing you a PID:
 - Your source code does not have any indication of being licensed under an open-source license. At a minimum, you need to check in a LICENSE file or similar to let users know how your code is licensed.
 - The 'org' field in your pid file should specify just the name of your org (eg, 'xfx'), not the whole path.
 - You appear to be using Arduino, which comes with its own PID. Can you clarify what you need a PID for?

PR #45, opened 2015-04-05
Applied for pid for OPQBox under openpowerquality

anthonyjchriste:
Requesting 1209/0987

### DIFF
--- a/1209/0987/index.md
+++ b/1209/0987/index.md
@@ -1,0 +1,8 @@
+---
+layout: pid
+title: OPQBox
+owner: openpowerquality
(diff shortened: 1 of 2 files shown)


Arachnid:
Thanks for your pull request! However, there are a couple of issues that need resolving before we can issue you a PID:
 - PIDs in the range 0x0000-0x0FFF are reserved for special uses, as described in http://pid.codes/1209/. Please choose a pid over 0x2000.
 - Your device doesn't appear to have a USB interface. Can you clarify what you need the PID for?

PR #47, opened 2015-04-05
add ii iigadget 1209:1111

hh:
Currently used on my custom yocto builds for Edison/UsbArmory/Pi, but will probably reused the same ID for Cyanogen and OpenWRT builds.

### DIFF
--- a/1209/1111/index.md
+++ b/1209/1111/index.md
@@ -1,0 +1,14 @@
+---
+layout: pid
+title: iigagdet
+owner: ii
+license: GPLv2
+site: http://instantinfrastructure.org
+source: https://github.com/ii/meta-edison-ii
+---
+
+This custom build for the edison includes keyboard, mass storage, and
+ethernet support. It's intended to be used to send keystrokes to PC bios
+on boot to enable booting from the mass storage and ethernet to set
+machines up from scratch.
+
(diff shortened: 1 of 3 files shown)


Arachnid:
Thanks for your pull request! However, there are a couple of issues preventing us from issuing you a PID:
 - PID codes in the range 0x1000 - 0x1FFF are reserved, as documented at http://pid.codes/1209/. Please pick a PID greater than 0x2000.
 - You appear to be using standard hardware which already has its own PID assigned. Can you clarify why you need a new PID?

hh:
While the repo supplied is for the edison, it's hoped we can use the same PID for standardizing our builds across all our supported hardware. When a supported ii device is connected to a usb host, we want to detect it differently than the hardware or vendor it was initially shipped as.

The various devices we want to support offer multiple ways to configure, reflash, and load drivers. We want to reduce that complexity.

Having a PID will allow us to detect devices already loaded with ii and assist those devices with the replication of themselves to other supported devices via flashing and re-configuring them.

Arachnid:
Thanks for the explanation. You still need to add a license to your source code repository. Other than that everything looks good.

hh:
I updated the repo to be a super repo that will include all supported builds and added a link to ii/meta-edison-ii
Both repos have a license now.
Thanks!

PR #50, opened 2015-04-05
"OpenVSM Virtual USB HID" PID request

Pugnator:
This PID is need for Virtual USB device. 
It will be used during simulation under Proteus for interracting with user's real hardware.

### DIFF
--- a/1209/C0DE/index.md
+++ b/1209/C0DE/index.md
@@ -1,7 +1,7 @@
 ---
 layout: pid
 title: OpenVSM Virtual USB
-owner: NonameGarage
+owner: Nonamegarage
 license: BSD
 site: https://github.com/Pugnator/openvsm
 source: https://github.com/Pugnator/openvsm/tree/testing
--- a/1209/C0DE/index.md
+++ b/1209/C0DE/index.md
@@ -1,0 +1,8 @@
+---
+layout: pid
+title: OpenVSM Virtual USB
+owner: NonameGarage
+license: BSD
+site: https://github.com/Pugnator/openvsm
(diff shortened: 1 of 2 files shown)


1209/C0DE/index.md
```diff
@@ -0,0 +1,8 @@
+---
+layout: pid
+title: OpenVSM Virtual USB
+owner: NonameGarage
```
Arachnid: This needs to have the same capitalisation as the directory you put your org file in.

Arachnid:
Thanks for your pull request! Can you clarify how this will be used? Will the user's operating system recognise this as a USB device? Will it have to load drivers? Would the test PID, http://pid.codes/1209/0001/ be suitable for this application?

Pugnator:
Thanks a lot for your reply!
The main idea of the project is that you can simulate IC by means of Lua script. Current default realization of USB stack available in Proteus is closed source one. Users will be able to see theirs virtual devices in devmgr and use it as it is a real one. And in this case yes, 0001 may be suitable but also I wanted to use it in several other projects our group is released, for example this openhardware/software GSM/GPS tracker for autosport I'm working on
https://github.com/Pugnator/ATRS
It will be nice to have the same PID everywhere. Sorry for my english, I hope I was able to clarify my intentions
I fixed capitalization issue by the way.

Arachnid:
Thanks for the details. It's generally a good idea to use a unique PID for each unique device that will be distributed to multiple users. As such, I'd recommend using our test PID 0x0001 for your virtual devices, and applying for a PID from us for anything that you expect to be distributed widely and require driver support.

PR #51, opened 2015-04-05
Add codes for KnightOS

SirCmpwn:
KnightOS acts as a generic hub and userspace applications can plug
themselves into it. We're reserving a code for the hub itself, as well
as for our MTP userspace service.


Arachnid:
Are you sure you wouldn't prefer `ADD5` or `CA1C`? ;)

SirCmpwn:
Those are excellent choices. We should totally use `CA1C`. Update incoming.

SirCmpwn:
Whoops! Can you fix that @Arachnid?

Arachnid:
They're displayed as {orgname} {devicename}. If you don't want the repetition, you can submit a PR that removes "KnightOS" from the device name.

PR #53, opened 2015-04-06
Adding Input Club PID for the Kiibohd Controller and Bootloader

haata:


Arachnid:
Are the devices you design open-source? If so, you should link to those in your PID requests.

haata:
Yes.
https://github.com/kiibohd/controller (currently does not have a license file because there are a few different licenses used).
https://github.com/kiibohd/kll
https://github.com/kiibohd/pcb
https://github.com/kiibohd/case

PR #60, opened 2015-04-08
Adding 1209:2342 for Big Red Button

andreas23:


Arachnid:
Thanks for your pull request!

I note your repository simply contains a copy of the Teensy USB code that you linked to, and no hardware designs.

Are you able to check in and OSS-license schematics and board layouts? If you're not using a custom PCB, can you explain your use-case for why you require a PID?

andreas23:
I've added a little more explanation to the web page. Detailed build instructions will follow, including schematics. I'm actually using a Teensy 2.0 instead of a custom PCB, because it fits the purpose very well. I still need a separate PID, since there will be software functionality specific to this application (e.g., programming the key sequence that is generated when the button is pressed).

Everything will be in the public domain.

PR #65, opened 2015-04-14
RobotMaker.club request VID 1209 PID EB01

robojay:
Hi!

We're working on a line of open source educational products.  The EB1 is our first one.  It's loosely based on the Adafruit Trinket (and others) with our own added features.

It's still in development but we want to solidify the driver side of things (with proper USB ids!) before releasing hardware into the wild.

Anyways - hopefully we meet the criteria!

Thanks,
Jay

### DIFF
--- a/1209/eb01/index.md
+++ b/1209/eb01/index.md
@@ -1,0 +1,8 @@
+---
+layout: pid
+title: EB1 电脑一号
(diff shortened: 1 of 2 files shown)


robojay:
Thanks!

PR #69, opened 2015-04-16
Request VID/PID

x893:


Arachnid:
Thanks for your pull request!

Please read carefully the guidelines at http://pid.codes/howto/. Specifically, your code and schematics need to be open-source licensed in order to obtain a PID code from pid.codes; your repository currently does not specify any license.

PR #77, opened 2015-04-16
VictorGrig USBscope PID pull request (updated)

VictorGrig:
Added to Github:

- source code
- burnable firmware
- EN/RU FAQ
- 2 executable files
- Altium Designer project
- USBscope PCB`s PDF-overview
- final PID pull request


Arachnid:
Hi,

I'm afraid you need to add these files to your own source code repository, independent of the pid.codes one, and send me a pull request that only changes the files as described in http://pid.codes/howto/. That is, the 'source' link in your pid file should point to a separate repository containing all your source.

PR #89, opened 2015-06-01
Adding request for ProteanSynth PID

mattomatto:


Arachnid:
Thanks for your pull request!

This looks mostly good, but I only see a PNG in the schematics directory. Are you able to include the original design files?

mattomatto:
It was created in Eagle and it might be a little complex to include all of the associated component libraries.  Would an image of the PCB layout suffice?

Arachnid:
Eagle files incorporate all the necessary footprints - it'll load for another user even without the libraries you used.

The point of releasing projects as Open Source is that people can modify, remix, and make it themselves; providing just a PNG doesn't make any of that possible.

mattomatto:
Ah ok no problem - I have updated with Eagle board and schematic files.
Thanks

PR #92, opened 2015-06-13
PID Code request

KK2345:
PID Code 1359 request


Arachnid:
Thanks for your pull request! Please read carefully the guide at http://pid.codes/howto/. Specifically, PIDs in the range 0x0000-0x1FFF are reserved; you need to request a PID that is not in that range.

KK2345:
Sorry, my mistake. Now it should be OK.

1209/2357/index.md
```diff
@@ -0,0 +1,8 @@
+---
+layout: pid
+title: TinyPassword
+owner: Karol Kucza
```
Arachnid: The value of the 'owner' field needs to match the name of your org directory, in this case `KarolKucza`. Otherwise, everything else looks good, though it'd be good if you could add a short description of what your device is after the metadata block!

KK2345:
I corrected it. Is it OK?

KK2345:
Thanks!

PR #94, opened 2015-06-27
Request PID BAD1 for CommLinkUSB project

gpoteau:
Request PID BAD1 for CommLinkUSB project


1209/BAD1/index.md
```diff
@@ -0,0 +1,9 @@
+---
+layout: BAD1
```
Arachnid: This needs to be "layout: pid".

gpoteau: Done.

PR #97, opened 2015-07-17
Added PAPCoDienTu org and FAB1 PID.

papcodientu:
Please kindly take a look into my request. Thank you!

--PAP


Arachnid:
I'm afraid I had to revert this because the PID file was not UTF-8 encoded. Can you please resubmit with UTF-8 encoded files?

PR #99, opened 2015-07-17
Changed org & PID files encoding to UTF-8.

papcodientu:
I changed PID file to UTF-8 encoded. Please kindly check and advice if any issue.

Thanks much,
PAP


Arachnid:
This doesn't seem to be synced with master; can you rebase from master?

PR #102, opened 2015-07-21
Request for new PID for sim controls interface

btodoroff:


org/SimAces/index.md
```diff
@@ -0,0 +1,5 @@
+---
+layout: org
+title: Brian Todoroff
```
Arachnid: Either this should be "Sim Aces", or your orgname (the directory this file is in) should be your own name.

Arachnid:
Can you please squash this into a single commit and re-push?

PR #113, opened 2015-09-29
Added Pinscape controller as 1209/EAEA

mjrgh:
This is an open source software project running on the Freescale KL25Z, which acts as a USB client device when running the project.

### DIFF
--- a/1209/EAEA/index.md
+++ b/1209/EAEA/index.md
@@ -1,6 +1,6 @@
 ---
 layout: pid
-title: Pinscape Controller
+title: Controller
 owner: Pinscape
 license: MIT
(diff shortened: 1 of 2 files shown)


1209/EAEA/index.md
```diff
@@ -0,0 +1,13 @@
+---
+layout: pid
+title: Pinscape Controller
```
Arachnid: FYI, as-is this will cause it to show up in the list as "Pinscape Pinscape Controller". Delete "Pinscape" here if you don't want that to be the case, or let me know if you want to go ahead and merge it as-is.

mjrgh: Thanks for catching that.  I just made the change you recommended.
Date: Tue, 29 Sep 2015 13:24:37 -0700

PR #116, opened 2015-10-31
3DRacers Pilot Board

3DRacers:
3DRacers Pilot is an Arduino-compatible board, similar to the Sparkfun
Pro Micro. 
A 3DRacers is a 3D printed RC car, that you can drive with an Arduino or a smartphone App.

The board schematics, sketch code and 3D models are on the
github organization repositories.

It needs 2 PID since the Caterina
bootloader runs on a different PID when activated, while the normal
program code runs on the main PID.

### DIFF
--- a/1209/FA57/index.md
+++ b/1209/FA57/index.md
@@ -1,0 +1,8 @@
+---
+layout: pid
+title: 3DRacers Pilot Board
+owner: 3DRacers
+license: Creative Commons Share-Alike 4.0
+site: http://www.3dracers.com/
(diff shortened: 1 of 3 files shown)


Arachnid:
Thanks for your pull request. This all looks good, but if I accept it as is, your entries will show up as "3DRacers 3DRacers ..." in the list. If you don't want that to be the case, you should remove "3DRacers" from the name field in the pid files.

3DRacers:
Thanks! I'll fix it right now :)

madarco:
Thanks!

PR #118, opened 2015-11-05
Add 'deshipu' org and request 'DE51' pid for it

deshipu:


Arachnid:
Thanks for your pull request!

Is this a project you expect other people will replicate on a wide scale? If not, you should be able to use your board's provided VID/PID, or select an 'internal' one for your own private use.

deshipu:
That's a good point. I don't expect this particular keyboard to be built by more people, as it's mostly manual work and thus hard to repeat. I treat it as a prototype. I do, however, have plans for a more repeatable design with the newer versions of this device. I suppose I will re-submit the pull request when that is ready.

Arachnid:
Hi Deshipu,

That sounds like a good idea. We're happy to issue people PIDs, but prefer to only do so when the device will be distributed to other users.

PR #123, opened 2015-12-11
Please add Global Boot pid of 0xB007. 

konsgn:
I'm trying to make a universal loader for various uC's based on the halfkay protocol. So far I have already made a functional bootloader for the PIC25K50 chips in assembly, so it fits withing the bootsector of that chip. Next attempt will be freescale or ST arm based processors.

### DIFF
--- a/org/Konsgn/index.md
+++ b/org/Konsgn/index.md
@@ -2,4 +2,3 @@
 layout: org
 title: Konsgn
 ---
-Just want a stable pid for a universal bootloader. Woo!
--- a/org/Konsgn/index.md
+++ b/org/Konsgn/index.md
@@ -1,0 +1,5 @@
+---
+layout: org
+title: Konsgn
+---
+Just want a stable pid for a universal bootloader. Woo!
--- a/1209/B007/index.md
+++ b/1209/B007/index.md
@@ -1,0 +1,9 @@
+---
+layout: pid
+title: Global_Boot
+owner: Konsgn
+license: GPLv2
+site: 
+source: https://github.com/konsgn/Global_Boot
+---
+A universal USB bootloader for microcontrollers based on the halfkay protocol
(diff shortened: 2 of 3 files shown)


Arachnid:
Thanks for your pull request! A few questions: 
 - Wouldn't it be more useful to have a separate PID for each type of device, so your bootloader can distinguish them and ensure it's loading the right kind of firmware?
 - Your boot images appear to be... actual images of boots. Where is the bootloader code?
 - The 'title' properties for org and pid should display the name of the org and device as you want them shown to users on the site - so it should probably be "Global Boot", not "Global_Boot". The orgname is prefixed to the device name, so don't duplicate it in the device name.
 - 'Site' is mandatory; if you don't have one, use the same URL as for your sourcecode repository.

konsgn:
Updated!
1) Not necessary, I'm going to identify device type based on serial number not vid/pid. 
2) Haha, Yup, just a fun startup image. I have added firmware too. The board design isn't really relevant to the goal of the project so I don't want to add that.
3) Updated org name to Konsgn. 
4) Done!

org/Konsgn/index.md
```diff
@@ -0,0 +1,5 @@
+---
+layout: org
+title: Konsgn
+---
+Just want a stable pid for a universal bootloader. Woo!
```
Arachnid: Better to leave this out if you don't have a good description for your org. Other than that, looks good!

konsgn:
Merp? 

Sooo remove the description underneath the title?

Arachnid:
Yes please, unless you have an org description that isn't about your desire for a PID. :)

Arachnid:
Talking about the org description here; the PID description is fine.

konsgn:
Haha, okay, done, it should be blank now.

konsgn:
Woot! thank you. now i just gotta make sure the new firmware works.... the last time I compiled it was months ago, so at least it's quite reliable.

PR #125, opened 2015-12-16
Fix typo

devanlai:
Sorry, I just noticed that the [license](http://pid.codes/1209/DA42/) doesn't show up because I accidentally injected a 3 into the license keyword when I was messing with my buffers.


devanlai:
Thanks for the quick response on both counts!

PR #126, opened 2015-12-21
Add GroupGets org and pid request for PureThermal 1

kekiefer:


Arachnid:
Thanks for your pull request!

Given this is a dev board, what do you need the USB PID for?

Can you please include a link to the design files in the PID file?

kekiefer:
Thanks @Arachnid, this is a great service.

The PureThermal 1 is only a dev board in the sense that it's set up to allow end users do development work for FLIR Lepton imagers host side. It is not an eval platform or anything like this, it is an end product.

I have updated the pull request with a link to the page with design docs.

Arachnid:
Thanks for updating the PR. I can't find source files anywhere (PCB design files and gerbers); are they available under an OSS license?

What do you anticipate using the PID for? If users primarily write their own code, you don't need a global PID; users will likely want their own if they're distributing anything. Or do you need it for the bootloader?

kekiefer:
We are not releasing the PCB design files at this time (though we likely will as the design matures and we have some visibility and recognition for the work). Forgive me if I've misunderstood your documentation, but it seems we have met the requirements:

> Containing __schematics or source code__ for a device with a USB interface
> ...
> If your project involves both hardware and software, __both need to be licensed under recognised OSS and OSHW licenses__.
> ...

We anticipate needing the PID to distribute drivers for the device; this is outside of the scope of the work we expect users to do (they just want to grab data and control the imager). It basically functions like a webcam, but in order to support additional vendor-specific bulk endpoints, we will need to associate a driver.

Arachnid:
As per the section you quote, "if your project involves both hardware and software, both need to be licensed under recognised OSS and OSHW licenses"; if you're not prepared to release the hardware source files, we're not able to issue you a PID.

mlyle:
@kekiefer the VUSB guys will give you a PID if you buy a software license from them (even the hobby one).  The license doesn't limit how you use the PIDs, only the VUSB firmware (which you don't need to use).

PR #128, opened 2016-01-14
Add PID for CCHS ESPlant board

projectgus:
Hello pid.codes crew!

Thank you for providing this extremely helpful service. Can we please have this PID for our OSHW project? :)

Angus
cc @mage0r

PR #129, opened 2016-01-19
Howto page: Note site field mandatory, can be the same as source field

projectgus:
Suggested small language edit for the "howto" page. When I was adding #128 I interpreted "may" as meaning the site field was optional.

Not a big change but should prevent someone following my mistaken reasoning again. :)

Thanks again for doing this, it's such a useful community service to provide!


Arachnid:
Thanks for the patch - this is much clearer.

PR #130, opened 2016-01-27
Add MickMad as organisation, reserve PIDs 813A-813B for HACK - HackAday Cortex Kit

MickMad:


Arachnid:
This looks fine, but be aware that the name shown on the pid listing page will be a concatenation of organisation and pid name, which will be very long in this case. Let me know if you want to go ahead as is.

MickMad:
Thanks for your prompt feedback, let me quickly review the pull request. I'll get back to you asap.

PR #131, opened 2016-01-27
Open source USB oscilloscope

elmot:


org/elmot/index.md
```diff
@@ -0,0 +1,5 @@
+---
+layout: org
+title: Elijah Motornyy(http://elmot.xyz)
```
Arachnid: You should leave the URL out of the title; it'll show up as a prefix to all PIDs under this org.

1209/e116/index.md
```diff
@@ -0,0 +1,8 @@
+---
+layout: pid
+title: Open-Source Oscilloscope based on STM32F3-DISCOVERY board
```
Arachnid: Does the project have a (shorter) name of its own?

Arachnid:
Can you describe what you need the PID for here? It looks like you're using stock hardware for the USB interface.

elmot:
Fixed

elmot:
Well, I use a stock prototype board, but it has only a physical  USB interface but no VID/PID. The on-board chip is shipped empty, firmware is completely mine, and, unfortunately, st.com does not provide any PIDs for products based on their chips or proto boards. And that's a reason why I request PID here.

elmot:
Thank you a lot.
great job!

PR #137, opened 2016-03-06
PID request for 1209/6464 for my GameCube to USB joystick mod

GGreenwood:
Hello! I would like to request a PID for one of my projects. It's an internal USB joystick conversion mod for the Nintendo Gamecube controller. The full source is licensed under MIT and can be found [here](https://github.com/GGreenwood/Shinewave). The firmware and hardware is found under `/firmware` and `/circuit`, respectively.

Thank you for providing this service!

Signed-off-by: Garrett Greenwood <garrettagreenwood@gmail.com>

### DIFF
--- a/1209/6464/index.md
+++ b/1209/6464/index.md
@@ -1,8 +1,9 @@
 ---
 layout: pid
 title: Shinewave
-owner: Electric Exploits
+owner: ElectricExploits
 license: MIT
 site: http://electricexploits.net/
 source: https://github.com/GGreenwood/Shinewave
 ---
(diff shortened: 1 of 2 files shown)


1209/6464/index.md
```diff
@@ -0,0 +1,8 @@
+---
+layout: pid
+title: Shinewave
+owner: Electric Exploits
```
Arachnid: Thanks for your PR! The Owner field needs to match the directory name of your org file. Other than that, this PR looks good.

GGreenwood:
@Arachnid 

Thanks for the quick turnaround! I've made changes as you suggested.

PR #139, opened 2016-04-23
Arduino Taiwan PID Request

will127534:


Arachnid:
Hi,

This PR looks fine, but are you sure you want to call it "Taiwan Arduino"? That's a pretty clear trademark violation.

will127534:
Will...I'll change to Taiwanduino 
I'm thoughtless when naming this board :P

PR #140, opened 2016-05-01
Added ManCave Made

uberflyx:
Requested one PID (DED1) for my open board the Quark One which is under the CERN OHL v1.2 open license and the bootloader + Arduino libraries which are under MIT license. Thanks

### DIFF
--- a/1209/DED1/index.md
+++ b/1209/DED1/index.md
@@ -1,0 +1,8 @@
+---
+layout: pid
+title: Quark One Serial
+owner: ManCaveMade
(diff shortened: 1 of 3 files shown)


Arachnid:
Thanks for the PR! Can you please amend it to point to the actual repositories? If there's multiple, perhaps there's a readme in one that references the others?

uberflyx:
Cool, I have changed it to point to my bootloader repo, which has a readme
that points to the hardware repo. Thanks!

PR #142, opened 2016-05-10
Duet 0.8.5 PID request

T3P3:
The hardware and firmware source for the Duet 0.8.5 are it two different Github repositories:
The hardware design is lead by Think3dprint3d (https://github.com/T3P3/Duet)
The firmware design lead by David Crocker (https://github.com/dc42/RepRapFirmware).

### DIFF
--- a/1209/D3D8/index.md
+++ b/1209/D3D8/index.md
@@ -1,0 +1,9 @@
+---
+layout: pid
+title: Duet 0.8.5
+owner: Think3dPrint3d
+license: MIT
+site: http://www.think3dprint3d.com/
+source: https://github.com/T3P3/Duet
+---
+Firmware source: https://github.com/dc42/RepRapFirmware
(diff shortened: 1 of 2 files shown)


1209/D3D8/index.md
```diff
@@ -0,0 +1,9 @@
+---
+layout: pid
+title: Duet 0.8.5
+owner: Think3dPrint3d
```
Arachnid: The owner string needs to match the directory name of the organization you created.

org/duet3d/index.md
```diff
@@ -0,0 +1,15 @@
+---
+layout: org
+title: Duet3d
+---
+The Duet 3dprinter controller is an advanced 32 bit controller for RepRap and other 3d printers,
```
Arachnid: This description text should probably be associated with the device, rather than the org.

Arachnid:
Apologies for the long delay.

PR #143, opened 2016-05-10
Request PID F12E for Feuermelder project

mibe:
The project is done, available under GPL v2 and is using the test PID 0x0001 ATM.


mibe:
Thanks! :+1:

PR #145, opened 2016-05-14
Request for PID - EE00 for SODA

Xplorer001:
SODA is SWD Open Source Debug Adapter for ARM Cortex M0, M3 and M4 chips.

### DIFF
--- a/1209/EE00/index.md
+++ b/1209/EE00/index.md
@@ -1,7 +1,7 @@
 ---
 layout: pid
 title: SODA(SWD OpenSource Debug Adapter)
-owner: Explore Embedded
+owner: ExploreEmbedded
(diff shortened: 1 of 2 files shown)


1209/EE00/index.md
```diff
@@ -0,0 +1,8 @@
+---
+layout: pid
+title: SODA(SWD OpenSource Debug Adapter)
+owner: Explore Embedded
```
Arachnid: This needs to be the same as the directory name for your org - in this case, 'ExploreEmbedded' without the space.

Xplorer001:
Thanks have updated the file.

Xplorer001:
Thank you!

PR #146, opened 2016-05-15
PID request for candleLight

HubertD:
* Add org/bytewerk.org
* Add 1209/2323

candleLight is a cheap USB CAN adapter with SocketCAN and WinUSB support

### DIFF
--- a/1209/2323/index.md
+++ b/1209/2323/index.md
@@ -1,0 +1,16 @@
+---
+layout: pid
+title: candleLight
+owner: bytewerk.org
+license: MIT, CERN OHL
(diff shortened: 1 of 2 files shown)


Arachnid:
Thanks for the pull request!

Is there source code for this project? If so, where can it be found, and how is it licensed? If not, for what do you need the PID?

HubertD:
As mentioned in the commited .md-files:
Hardware design: https://github.com/HubertD/candleLight (KiCAD, CERN OHL)
Firmware source: https://github.com/HubertD/candleLight_fw (MIT license)

PR #148, opened 2016-05-22
kuku.eu.org keyboard PID registration

anszom:


Arachnid:
This looks fine, but note that the title of the device in the registry will be the concatenation of org and title strings - "Andrzej Szombierski (kuku.eu.org) kuku.eu.org keyboard", which seems a bit verbose.

anszom:
You're right. I've removed the domain from the org name, keeping it only for the device (the device will report this full name in its descriptor strings).

I'm not yet familiar with github's pull requests, do I need to resubmit something, or is the pull request updated automatically?

PR #149, opened 2016-05-23
add SatoshiLabs and TREZOR

prusnak:


Arachnid:
Small world!

Surely Trezor already has a PID, though?

prusnak:
On 23/05/16 17:27, Nick Johnson wrote:
> Small world!
> 
> Surely Trezor already has a PID, though?

Yes. But it was not "officially" assigned though. Also this is a part of
upcoming TREZOR v2 preparations.

-- 
Best Regards / S pozdravom,

Pavol Rusnak <stick@gk2.sk>

Arachnid:
Righto. If this ends up being used for the Trezor 2 instead, please send a PR at the time to update the registry!

PR #151, opened 2016-06-02
SeregaSPb IT-specialist 

SeregaSPb:
USB flash drive for easy system repairment, fast OS instalation, setup and protection from rootkits, BadUSB and other hacks and treats. 
It is supposed to use a modified firmware of the USB controller with non-writable CD-ROM and hidden secure partitions and more.


Arachnid:
Thanks for your pull request!

Unfortunately, your repository does not contain any source code or schematics for the device, so we can't issue you a PID. Devices must be open-source hardware and software to be eligibile for a pid.codes PID.

When resubmitting, please keep the title and description of the entries relevant to users reading the PID entry at a later stage; comments about why you need the PID should go in the pull request, not in the entries themselves.

PR #158, opened 2016-07-27
add pid 6666 for vsflogic

talpachen:


Arachnid:
Thanks for your pull request!

Your repository seems to be a software project; can you expand on why you need a PID, and what hardware it will be used on?

talpachen:
Thanks!

Vsf is an embedded event-driven framework. It can be used on many MCUs, see here: https://github.com/talpachen/vsf/tree/vsflogic/vsf/interfaces/cpu

I want to DIY low cost Labtool board based on LPC-link2. An USB-based, 11-channel logic analyzer with up to 100MHz sampling rate, and with 2 additional analog channels.

Arachnid:
Thanks for the details. Are you combining existing firmware with existing hardware? Do you plan to distribute this widely? I'm trying to get a sense for why you need the PID.

talpachen:
Yes, Although the official firmware has its own PID, but is not compatible with my protocol. I need a new PID for my firmware.

PR #162, opened 2016-08-05
Request PID for RomanStepanov Shifter/Pedals Adapter

robotsrulz:
Please provide a desired PID number for Logitech G25/G27 standalone USB HID adapter. Thanks!

### DIFF
--- a/1209/F00D/index.md
+++ b/1209/F00D/index.md
@@ -1,7 +1,7 @@
 ---
 layout: pid
 title: Shifter/Pedals Adapter 
-owner: RomanStepanov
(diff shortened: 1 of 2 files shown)


1209/F00D/index.md
```diff
@@ -0,0 +1,8 @@
+---
+layout: pid
+title: Shifter/Pedals Adapter 
+owner: RomanStepanov
```
Arachnid: Please set this to the directory name of the org file - in this case, 'robotsrulz'.

robotsrulz:
Does it look correct now?

PR #168, opened 2016-08-24
Create 8086

Misfittech:
Commit for the arduino zero bootloader PID


Arachnid:
Hi,

Can you please explain more about your planned use?

PR #170, opened 2016-08-24
Misfittech patch 1

Misfittech:


Arachnid:
Thanks for the updated PR. You still need to provide hardware source files in a manner that would permit someone to create a derived work, however: gerbers at a very minimum, preferably source files.

If you're requesting separate PIDs for bootloader and device, please include 'bootloader' in the title of the appropriate PID file.

PR #177, opened 2016-09-08
Request for 1209:AA00

SergOskin:


1209/AA00/index.md
```diff
@@ -0,0 +1,8 @@
+---
+layout: pid
+title: LinuxCNC HID Extender
+owner: Serg Oskin
```
peternewman: This needs to say "serg" @SergOskin to match the folder org/serg/index.md.

mlyle:
Note there's no hardware or software implementation at the target repository from what I can see.. Typically thigns have not gotten merged when this is the case.

SergOskin:
peternewman,
I hope I understand that it is necessary to fix...

mlyle,
The project has just been born today. The first version of the software will appear in about a week. By this time, it would be desirable to have a VID: PID. The first version of Hardwar - STM32 demoboard.

peternewman:
That's done the job thanks @SergOskin .

In terms of testing, there is a test PID you could use. http://pid.codes/1209/0001/

Arachnid:
Hi Serg,

As others have pointed out, you need to have actual hardware and software before we'll issue you a PID - we don't issue them on a speculative basis. Please feel free to use the test PID until you're ready to make a batch of hardware for distribution to other users, at which point feel free to submit a new PR.

PR #178, opened 2016-09-09
More data cleanse

peternewman:


Arachnid:
Thanks for spotting these.

PR #180, opened 2016-09-09
Escape the colons in the title

peternewman:


Arachnid:
Thanks again! Any interest in becoming a co-maintainer?

peternewman:
Thanks for the offer @Arachnid but I think I'd rather concentrate on the existing open source stuff I'm involved in. Although by the time I've finished, your life should be easier, and the issues I've been fixing should be caught by a Travis job before they try and merge them.

My final end game, and the reason I'm doing this, is to get the data in a state where it can be used to update http://www.linux-usb.org/usb.ids and hence can be used in stuff like Wireshark https://github.com/wireshark/wireshark/blob/master/epan/dissectors/usb.c . Then I'll stop bothering you with all these PRs :).

Arachnid:
> the issues I've been fixing should be caught by a Travis job before they try and merge them.

Yes, I've been meaning to set up CI, which would save me a lot of trouble with mundane issues (and prevent those you've identified as slipping through the cracks)

> My final end game, and the reason I'm doing this, is to get the data in a state where it can be used to update http://www.linux-usb.org/usb.ids

Oh, that would be really excellent. Are you in a position to update the PID listing, too? Every time I tried submitting, the automated system rejected my diff, even though it applied cleanly when I tried it locally.

peternewman:
I was assuming the automated system would just work. Have you already got code to convert pid.codes into usb.ids then?

I generated a copy of the linux-usb website code here https://github.com/peternewman/usbids-original from their diff and the PCI IDs site. I think there was some stuff in their code suggesting you actually needed to register first now.

Arachnid:
> I was assuming the automated system would just work. Have you already got code to convert pid.codes into usb.ids then?

No, I was just talking about updating the single PID value, which is still set to its old owner. I tried composing a diff by hand, and while it applied locally, the email submission always failed.

PR #183, opened 2016-09-28
Enable initial Travis CI workflow

peternewman:
And fix the resulting errors!

You'll need to turn on Travis on your repo to get the benefits, I also highly recommend the protected branch settings on GitHub, even for admins! (You can always temporarily override them as admin if you need to).

Some things to note:

- It's not checking for unknown tags, as you're using header, article and footer in the HTML, if they were swapped for a valid tag like div or span, we could enable that check too.
- I think the link validation will catch people giving their org folder the wrong name, but I still need to check

What do you want to do about external links? There are a handful of dead ones currently, hence the allowed failures section on Travis.


index.html
```diff
@@ -16,15 +16,15 @@
       <div class="post">
         <a href="{{ post.url | prepend: site.baseurl }}" class="post-link">
           <p class="post-meta">{{ post.date | date: "%b %-d, %Y" }}</p>
-          <h3 class="h2 post-title">{{ post.title }}</h3>
+          <h3 class="h2 post-title">{{ post.title | xml_escape }}</h3>
         </a>
-        <p class="post-summary">
+        <div class="post-summary">
```
peternewman: Hopefully this doesn't break the look, but htmlproofer complains about the closing /p.

Arachnid:
Oh wow, amazing job - thank you!

RE external links, it'd be good to catch people submitting invalid URLs in their submissions, but I definitely don't want to block one person's PR because another's has gone 404. In the absence of some way to do that, I'd err on the side of not validating them.

pids.md
```diff
@@ -7,6 +7,6 @@ permalink: /pids/
 {% for vid in vids %}
   {% assign urlparts = vid.url|split:"/" %}
   {% assign vidno = urlparts[1] %}
-  <h3><a href="{{vid.url}}">0x{{urlparts[1]}}</a></h3>
+  <h3><a href="{{vid.url}}">0x{{vidno}}</a></h3>
```
peternewman: This line is still generating a pair of p /p around it, which the validator doesn't like. Any thoughts on how to get rid of it @Arachnid ? I wondered if switching the whole line to markdown might work?

Arachnid: Hm. I suspect it's the empty lines from the assign statements, which markdown is treating as empty paragraphs. Maybe [put it inside a div](http://stackoverflow.com/questions/701042/disable-markdown-for-a-block) or try moving them all onto one line?

peternewman:
No worries.

I'll have a ponder about external links. There is some caching stuff regarding them and I think you can ignore URLs, so you could ignore the broken ones; it's not perfect, but may be an acceptable compromise.

Alternatively you can manually refer to the allowed failures section of Travis.

peternewman:
Can you click the button here to turn on Travis please @Arachnid https://travis-ci.org/pidcodes/pidcodes.github.com/ ?

Arachnid:
Of course. Sorry, I thought that was only useful after submitting this PR.

peternewman:
I hoped it would, and indeed after a new commit it's started showing the PR builds: https://travis-ci.org/pidcodes/pidcodes.github.com/builds/164354824 !

pids.md
peternewman: I manage to fix it by just removing the leading spaces.

index.html
peternewman: Let me know if it does @Arachnid and I'll do something else instead.

peternewman:
Given this is now green, I'd be tempted to merge it now if you're happy, and I can look at addressing the other bits going forward, but in the meantime you'll already get the benefits of some initial Travis testing being in place.

Arachnid:
Indeed I am - thanks again for all the hard work!

index.html
peternewman: Looks fine still to me.

Issue #184, opened 2016-10-07
Travis TODO


peternewman:
- [ ] Check Open Graph via html-proofer https://github.com/gjtorikian/html-proofer/issues/111
- [ ] Check tags when html-proofer supports HTML 5
- [ ] Check hex in markdown including case
- [ ] Fail on broken external URLs

Arachnid:
FYI, https://github.com/pidcodes/pidcodes.github.com/pull/186 is passing despite the owner field being incorrect.

peternewman:
It did actually catch it, but as an external link, which we currently ignore, unless you manually review them.

Before:
https://travis-ci.org/pidcodes/pidcodes.github.com/jobs/167023698#L1651

After:
https://travis-ci.org/pidcodes/pidcodes.github.com/jobs/167102370#L1572

peternewman:
Still not right, see the last two commits here:
https://github.com/pidcodes/pidcodes.github.com/pull/195

peternewman:
I've realised, it's not catching the broken org links, as the site is already coded to defend against them.

orgpath contains the actual, potentially incorrect URL https://github.com/pidcodes/pidcodes.github.com/blob/master/_layouts/pid.html#L2
But the link used on the page is org.url, which comes from https://github.com/pidcodes/pidcodes.github.com/blob/master/_layouts/pid.html#L3 and only matches if the two are the same already, otherwise you'll get nothing for org.url or org.title, but there will be nothing for the proofer to find either.

peternewman:
Proved here:
https://travis-ci.org/peternewman/pidcodes.github.com/jobs/179990067

Are you happy for me to make the change @Arachnid ? It would only make a difference if something gets merged with a broken link.

Arachnid:
Good catch - please do!

peternewman:
Done, see https://github.com/pidcodes/pidcodes.github.com/issues/198 . To be undone once we get some better suited method of checking the fields, like the hex markdown checks will require.

peternewman:
Confirmed the link validation will catch people giving their org folder the wrong name, sorted in #198

peternewman:
Some possible validation options:
https://github.com/cityoffortworth/jekyll-data_validation
https://github.com/18F/jekyll_frontmatter_tests

Arachnid:
I'm reasonably happy with verifying the generated HTML; is there a big disadvantage to doing that, or stuff we can't catch?

[...]

PR #185, opened 2016-10-07
Check Open Graph via html-proofer

peternewman:


peternewman:
Progress on #184 .

peternewman:
There may be one edge case with this, the checks may fail for a PR against a new PID, as the "external" URL in the Open Graph won't be valid, as the PR hasn't been approved yet. I might need to flag it as an internal link or something.

Arachnid:
Sorry, just saw this; reviewing now.

peternewman:
My edge case prediction was correct, but as we currently don't care about external links, it doesn't (yet) break things, making it internal should fix it:
https://travis-ci.org/pidcodes/pidcodes.github.com/jobs/167023698#L1640

PR #187, opened 2016-10-13
Request for Explore M3 DFU and Virtual COM port PID

Xplorer001:
We need 2 PIDs for our open hardware and Software project Explore M3. The device works in DFU mode for programming and comes up as virtual COM port when the application runs. Two PIDs are required to avoid driver conflict on some of of the operating systems like windows 10. Thanks!


Arachnid:
Thanks for your PR! Please be sure to include a space between `link:` and the URI so it parses properly.

Xplorer001:
Changed the markup, please check now! Thanks.

PR #190, opened 2016-10-19
Make pid.codes an internal domain for htmlproofer

peternewman:


peternewman:
This needs some work still, or possibly a fix on the htmlproofer side.

Arachnid:
Should this be committed, then?

peternewman:
I think not until I've fixed the issues on the htmlproofer side thanks.

peternewman:
See https://github.com/gjtorikian/html-proofer/pull/361

peternewman:
This should be good to merge now @Arachnid .

PR #191, opened 2016-10-25
VitaSound USB MIDI Key

UA3MQJ:


Arachnid:
Thanks for your pull request!

I don't see any hardware design files checked into the repository. If this is for a development board, can you not use the VID/PID provided with the board, or the pid.codes experimental PID?

UA3MQJ:
u

PR #195, opened 2016-11-05
Requested PID for PancadariaStick.

streeto:


Arachnid:
Thanks for your PR!

Can you please change your org name to something reflecting who you are, rather than just this product?

If there's hardware, you need to include source files for that in the source repository.

streeto:
Changed the org to StreetoArcade. Also changed the site, now it points to another repo which hosts a v-usb atmega8 version. This one has the code and hardware design files. Both versions, LUFA and v-usb, implement the same USB descriptor, and I intend to use the same PID for both. Hope this is OK.

Thanks!

PR #204, opened 2016-12-23
Add a site field to orgs

peternewman:
It currently looks like this:
https://peternewman.github.io/pidcodes.github.com/org/openlightingproject/

Happy to restyle it, but I figured while I'm making links clickable, I may as well turn them into metadata too, so it can be used in other ways in future.


peternewman:
This can go in now @Arachnid before #200 which I'll then tweak to use more of this.

Arachnid:
Another useful addition, thanks!

peternewman:
Thanks @Arachnid . I'm considering doing similar for some additional PID fields too, although I'm not entirely sure how granular to be.

PR #205, opened 2016-12-24
Request PID for Temp monitor

nilssonaxel:


Arachnid:
Hi,

Do you plan to manufacture and distribute this device? If not, you might consider just using our freely available test PID.

nilssonaxel:
Hello!

We plan on manufacturing and distributing the device.

PR #217, opened 2017-02-04
PID request

arzaman:
PID request for open source SmartPID project

### DIFF
--- a/1209/4646/index.md
+++ b/1209/4646/index.md
@@ -1,6 +1,6 @@
 ---
 layout: pid
-title: SmartPID SPC1000
+title: SPC1000
 owner: SmartPID
 license: GPL V3
 site: http://www.smartpid.com/
--- a/1209/4646/index.md
+++ b/1209/4646/index.md
@@ -1,0 +1,8 @@
+---
+layout: pid
+title: SmartPID SPC1000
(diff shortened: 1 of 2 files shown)


1209/4646/index.md
```diff
@@ -0,0 +1,8 @@
+---
+layout: pid
+title: SmartPID SPC1000
```
Arachnid: Since the org name is prefixed to the title, this will show as 'SmartPID SmartPID SPC1000'; is that what you intended?

1209/4646/index.md
```diff
@@ -0,0 +1,8 @@
...
+owner: SmartPID
+license: GPL V3
+site: http://www.smartpid.com/
+source: https://github.com/arzaman/smartPID
```
Arachnid: I don't see any hardware design files here, only schematic PDFs. Can you include hardware source files in your repository, please?

1209/4646/index.md
arzaman: understood
would prefer "SmartPID SPC1000" so better to remove SmartPID from the tile

1209/4646/index.md
arzaman: what format?

PR #218, opened 2017-02-04
PID request for Danyboard M0

Gius-8:
Request PID for danyboard m0 bootloader. Thanks :)

### DIFF
--- a/1209/812A/index.md
+++ b/1209/812A/index.md
@@ -1,0 +1,8 @@
+---
+layout: pid
+title: Danyboard M0
+owner: Danyboard
+license: GPL
+site: http://www.danyboard.com/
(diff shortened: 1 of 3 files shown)


1209/812A/index.md
```diff
@@ -0,0 +1,8 @@
+---
+layout: pid
+title: Danyboard M0
+owner: Danyboard
+license: GPL
+site: http://www.danyboard.com/
+source: https://github.com/Gius-8/Danyboard_M0/
+---
```
Arachnid: Please adjust the title on this and the other PID to indicate which one is for the bootloader and which is for the client software.

Gius-8: Hi, I changed the title of the PID 8123 with Danyboard M0 bootloader. Thanks :)

PR #228, opened 2017-02-28
add ODrive v3.2

madcowswe:


Arachnid:
Thanks for your pull request! And this is an awesome project - I saw it on HaD and thought it was brilliant.

Do you have hardware designs, or are you using a standard development board? If the latter, do you plan to distribute this to people? What do you need the PID code for?

madcowswe:
The hardware design is here: https://github.com/madcowswe/ODriveHardware

madcowswe:
@Arachnid Anything else that you need?

madcowswe:
By the way, thanks for running this service, it's great to have access to unique codes for OSHW without having to pay huge fees aimed at large companies ;D

PR #232, opened 2017-03-04
Try making liquid strict

peternewman:
I'm hoping this might help the validate layout options issues in #184.

I've not tested what this looks like yet or if my code changes break everything, but the number of links checked is still the same, which can only be a good thing.


peternewman:
The website looks fine still so I don't think I've broken it!

PR #236, opened 2017-03-27
PID for GBCartFlasher

melka:
Hi,

This is my pull request for a VID / PID pair for a USB gameboy cart flasher. The project was already open sourced by the original developers, but I had to make a couple of changes in order to get it to work easily on MacOS, and one of those changes was to change the VID / PID on an FTDI FT232RL in order to stop MacOS to try and load the default kexts and being able to use libftdi.

Hope you'll accept the request.

Regards,

### DIFF
--- a/org/blaste/index.md
+++ b/org/blaste/index.md
@@ -1,0 +1,6 @@
+---
+layout: org
+title: blaste
+site: http://www.blaste.net (in progress)
+---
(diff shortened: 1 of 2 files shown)


org/blaste/index.md
```diff
@@ -0,0 +1,6 @@
+---
+layout: org
+title: blaste
+site: http://www.blaste.net (in progress)
```
peternewman: You'll need to move the (in progress) comment elsewhere, as it confuses the system and tries to put it in the URL.

PR #244, opened 2017-04-04
Added PD Buddy organization and PID 9DB5

Ratfink:
1209:9DB5 - PD Buddy Sink

PD Buddy Sink is a device I'm making that makes USB Power Delivery easy to use for any electronics hobbyist.  The user configures the Sink with a particular voltage and current, then the Sink requests the user setting from the power supply whenever it's plugged in.  The configuration is done over a USB CDC-ACM interface, so having a non-testing PID for it would be wonderful.

Thanks,
—Clay Hobbs

### DIFF
--- a/1209/9DB5/index.md
+++ b/1209/9DB5/index.md
@@ -1,0 +1,10 @@
+---
+layout: pid
+title: Sink
+owner: pd-buddy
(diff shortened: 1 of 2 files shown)


Ratfink:
Thanks!

PR #245, opened 2017-04-08
PID request for Intelectron Framework

Intelectron:
Pull Request for the following _VID_/**PID** pairs

[_1209_/**a8b0**](https://github.com/Intelectron/pidcodes.github.com/blob/master/1209/a8b0/index.md)
[_1209_/**a8b1**](https://github.com/Intelectron/pidcodes.github.com/blob/master/1209/a8b1/index.md)


Intelectron:
@peternewman , We've tried but due to some unknown reason, We arn't not being able to remove it (^M DOS ending).
Please suggest a workaround if possible.
Regards,
Intelectron Team

uriba107:
Your can either ensure your git client is not configured to push using CRLF
https://help.github.com/articles/dealing-with-line-endings/#platform-all

or you can just edit the files and remove it
Most fail proof way is on linux, open file in Nano, and save changes, when saving, you would see "(Dos line ending)" hit ALT+M to "disable" it and use unix, then save.

in windows just use a proper editor (notepad++, atom or whatever) - they all have the option to change line endings.

Intelectron:
Thank you for pointing the issue @peternewman,
Thank you for suggesting the solution@uriba107
We think it's fixed now. Can you verify ?

PR #269, opened 2017-08-24
Request for ADDA for MicroPython

nickzoic:
So my hardware hasn't actually arrived yet and my schematic is drawn on the back of an envelope, but ADDA would be such a nice PID for a MicroPython board :-).

### DIFF
--- a/1209/ADDA/index.md
+++ b/1209/ADDA/index.md
@@ -1,0 +1,11 @@
+---
+layout: pid
+title: MicroPython Boards
+owner: micropython
+license: MIT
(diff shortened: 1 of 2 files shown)


1209/ADDA/index.md
```diff
@@ -0,0 +1,11 @@
+---
+layout: pid
+title: MicroPython Boards
+owner: micropython
```
peternewman: The capitalisation of this needs to match the org folder name for the other file, i.e. MicroPython (or rename the folder).

PR #284, opened 2017-11-09
Update ODriveRobotics website, add ODrive v3.3 and v3.4

madcowswe:
Thanks!


Arachnid:
Do you need new PIDs for these new revisions? Are their interfaces incompatible with older versions?

madcowswe:
They are not strictly required, only nice to have. If you think it's unfair to allocate the revision range, I can change the description to refer to ODrive v3 in general.

madcowswe:
Okay, I have changed the PR to simply update the info, not allocating any new codes.

PR #287, opened 2017-11-26
Croduino SAMD PID request

davidzovko:


davidzovko:
Hello

sorry to bother - but is this still active, is it still possible to assign new PID? 

Thank you!

peternewman:
@Arachnid how come this was allowed through when it's in the reserved range:

> PIDs in the range 0x1000 - 0x1FFF are reserved for devices produced by InterBiometrics, the original owner of this VID.

Arachnid:
@peternewman Accepting it was a mistake; unfortunately there's no way to undo an allocation.

PR #293, opened 2018-01-05
PID request for Monero Hardware firmware(s) and one bootloader (#293)

michaesc:
The two (instead of one) firmware PIDs map to two different device types (with different architectures) we plan to release.

### DIFF
--- a/1209/B0B0/index.md
+++ b/1209/B0B0/index.md
@@ -1,0 +1,9 @@
+---
+layout: pid
+title: Monero Bootloader
+owner: Monero Hardware
+license: CERN
+site: https://getmonero.org/
+source: https://github.com/monero-project/kastelo/
+---
+The Monero Hardware Wallet, a Open Source and secure cryptodevice
--- a/1209/C0DA/index.md
+++ b/1209/C0DA/index.md
@@ -1,0 +1,9 @@
+---
+layout: pid
+title: Monero Firmware
+owner: Monero Hardware
+license: CERN
+site: https://getmonero.org/
+source: https://github.com/monero-project/kastelo/
+---
+The Monero Hardware Wallet, a Open Source and secure cryptodevice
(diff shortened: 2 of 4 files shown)


Arachnid:
Hi,

Thanks for your pull request!

Unfortunately, the repository you link to is 404 (or restricted access). We can't issue PIDs to you until you've published open-source software and hardware designs for your project. In the meantime, you're very welcome to use the reserved development PIDs.

michaesc:
Hello @Arachnid, the reason you get a 404 reply is that our repository has not been renamed yet from the upstream maintainer. If you see the bug report referenced in this pull request 'Customisations requested to Sekura repository' it states that the repository named [monero-project/sekura](https://github.com/monero-project/sekura/) will be renamed to **monero-project/kastelo** some time soon.

1) That's why I hope you'll decide in our favour, after seeing that we have published open-source software and hardware designs for our project (see the live repository link above.)

2) If you prefer, I'll change the pull request to refer to the deprecated _sekura_ repository name. After the upstream (Monero Project) maintainers correct our repository name, I'll send you a new pull request to correct the name to _kastelo_ here as well.

I prefer the short cut method number 1, because I know there is no opposition upstream to change our repository name. In fact, we're not allowed to use the name _sekura_ anymore.

michaesc:
Hello again @Arachnid,

After a succesful time at FOSDEM and several publications of websites, project management, hardware design, and software implementations, we're still holding off on using 1209/B0B0 and 1209/C0DA until you accept our request. If you'd like to see the Opensource code and location where these things are published:

https://github.com/monero-project/kastelo/blob/master/firmware/monezor/src/usb.c (search 'idVendor')
https://github.com/m2049r/monerujo-hw/blob/master/src/usb.c (search 'idVendor')

...those are two (!) example of firmware we'd like to use a 1209 assignment in, both Opensource and publicly available.

Would you like for us to make changes, or are we okay?

PR #309, opened 2018-03-27
Create index.md

Portwell:
I creat a new Arduino base board.


Arachnid:
Please create one PR containing both your new organisation and any PIDs you are requesting.

PR #326, opened 2018-04-28
USB DFU Bootloader for SAMDx1

majbthrd:
lightweight (1kB rather than existing 4kB or more) USB bootloader for the Atmel/Microchip SAMD11 and SAMD21 microcontrollers conforming to the USB DFU protocol


majbthrd:
adding second request for Dapper Miser, a minimalist CMSIS-DAP debug unit that provides ARM debugging at a very low implementation cost

PR #331, opened 2018-05-21
PID request for Meow Meow

sabas1080:
Hi 
I want to request a PID for the Meow Meow hardware

https://github.com/ElectronicCats/MeowMeow

with Open Source Hardware Certificate UID : MX000003

http://certificate.oshwa.org/certification-directory/

and and add as an organization to Electronic Cats

www.electroniccats.com

Thanks

### DIFF
--- a/1209/BAB1/index.md
+++ b/1209/BAB1/index.md
@@ -1,0 +1,9 @@
+---
+layout: pid
+title: CatSat Zero
+owner: Electronic Cats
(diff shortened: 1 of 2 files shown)


1209/BAB1/index.md
```diff
@@ -0,0 +1,9 @@
+---
+layout: pid
+title: CatSat Zero
+owner: Electronic Cats
```
Arachnid: This needs to match the directory name of the org file.

PR #332, opened 2018-05-22
Org for my Maestro boards

DavidDGrande:
Maestro boards are AVR microcontrollers


Arachnid:
Please see the docs on pid.codes; the orgfile needs to be submitted as part of a PR for the first PID, and needs to match the org used in the PID file.

PR #336, opened 2018-05-27
STM32 USB Serial scope

arpruss:
Thanks!


1209/AAAA/index.md
```diff
@@ -0,0 +1,8 @@
+---
+layout: pid
+title: STM32 USB Serial Oscilloscope
+owner: Omega Centauri Software
```
Arachnid: This needs to match the name of the organisation folder.

arpruss: 1. There is no PCB design. If all one wants is to monitor TTL stuff, it should eventually work with a standard blue or black pill STM32F103 board with an F303 soldered on it.

2. Do you mean that I should remove the spaces from the owner name, to match the OmegaCentauriSoftware folder?

Arachnid: Yes, the 'owner' field needs to be the directory name of your organisation.

arpruss: OK, fixed.

Arachnid:
Please open a new PR when your repository has hardware and software designs and an open-source license.

PR #343, opened 2018-07-10
added ArduPilot PID request

tridge:


tridge:
there seems to be a problem with the travis Ruby setup. I've tried triggering CI a couple of times with no luck

tridge:
I have added a patch updating the ruby version used for travis checking, which allows travis to pass again

Arachnid:
Thank you!

PR #347, opened 2018-08-02
Add Nerves-Project Org and first device

GregMefford:


Arachnid:
Are hardware designs for the 'ethernet + serial gadget' available anywhere? If so, you need to link to them in the PR.

GregMefford:
This project is based on a Raspberry Pi Zero, so the hardware designs aren’t open-source, but it is a commodity item that anyone could purchase.

Arachnid:
Can you clarify what the PID is for, then?

GregMefford:
When creating a software-defined “USB Gadget” device using Linux, you need to specify a USB VID/PID, just the same as you would in a hardware design with a USB host chip. Most people seem to just “use one that happens to work,” but I would like for us to have a dedicated PID to specify that it’s a Nerves-based serial + Ethernet composite device.

PR #348, opened 2018-08-11
add org "dm9records" & PID request

hsgw:


hsgw:
I got ids other way. thanks.

PR #355, opened 2018-08-28
request for org X41 and pid 5350 for SpacePort

X41:


Arachnid:
Is this BSD, or CC? Your header says one thing, but your source says another.

X41:
Oh it should be CC. Where does it say BSD? i couldn't find that anywhere

(i should probably note, that the hardware is mine and released under CC. the software is just a fork that's licensed under MIT. should i add both in the request or just the one?)

PR #367, opened 2018-10-26
Request pid for asceension

hidsh:
Hello pidcodes,

I would like to get PID for my new USB keyboard which is now making.

Thank you.


hidsh:
Thanks to merge me!

PR #374, opened 2018-11-08
added Thunkit Electronics PID 2424

cmccaskey:
I would like this PID so that I can use Gemma boot from Adafruit on my own board using the ATtiny85

### DIFF
--- a/1209/2424/index.md
+++ b/1209/2424/index.md
@@ -1,0 +1,8 @@
+---
+layout: pid
+title: T85DUINO
+owner: Thunkit Electronics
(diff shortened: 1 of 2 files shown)


1209/2424/index.md
```diff
@@ -0,0 +1,8 @@
+---
+layout: pid
+title: T85DUINO
+owner: Thunkit Electronics
```
peternewman: This should be the folder name of the org page:
```suggestion
owner: ThunkitElectronics
```

PR #384, opened 2018-12-09
Removing ORG

Intelectron:
Removing this ORG as all rights will be transferred to the original owner([Sandeepan](http://pid.codes/org/Sandeepan/))

Also, please remove association to the assigned VID/PID pairs and let them be available for everyone to use

1. 1209/a8b0
2. 1209/a8b1


Arachnid:
This PR *also* adds two new PIDs, which is not what it's described as doing.

PR #386, opened 2018-12-12
Request pid 0x2440 for the GnuPG Verein.

dd9jn:
We like to hand out tokens as well as giving everyone an opportunity to use this hardware and software for their own idea on how to implement an OpenPGP compliant token.  For interoperability it would be good to have a dedicated VID/PID for such small series of tokens.  After all the different implementations can apply for a free manufacturer ID (according to the OpenPGP card specs) and use this to distinguish their implementations.  Due to the usual problems the VID of the FSIJ can't be used for own experiments or tokens and thus the missing VID/PID is  major hurdle to deploy such tokens.

FWIW, 2440 is the original RFC number for OpenPGP.  4880 being the current one but I hesitate to ask for two PIDs just to reserve it.

### DIFF
--- a/1209/2440/index.md
+++ b/1209/2440/index.md
@@ -1,0 +1,12 @@
+---
+layout: pid
+title: An OpenPGP Token
+owner: GnuPG-Verein
+license: GPLv3, CC BY-SA 3.0
+site: https://gnupg.org/verein/token.html
+source: https://gnupg.org/verein/token.html
+---
+
+A crypto token for the members of the Verein (with the Verein's
+OpenPGP manufacturer id) and others (using other manufacturer ids or
+the random manufacturer id range).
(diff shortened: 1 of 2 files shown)


Arachnid:
Thanks for your pull request! Is the project at the point where you need a dedicated PID? If it's still in the prototyping phase, you're welcome to use our test PID.

dd9jn:
Well, the hardware is in actual use but several people asked how to build or modify the hardware on their own without running into problems with the VID:PID.  The VID:PID in use when you buy ready tokens from the author, the FSF, or Seeed Studio can't be used if you want to distribute the tokens yourself.  

Having a dedicated VID:PID for these types of tokens would be very useful so that they can be made part of the Linux distros USB VID:PID registries.  The need for the VID:PID is merely so that it is possible for GnuPG to access those tokes without requireing the user to fiddle with the USB setup or worse use arbitrary or even 0000:0000 ids.

For our own project we need to decide on a case and can then start to flash and distribute the tokens.

PR #387, opened 2018-12-19
Setup obfuscated organisation and request a pidcode for thekbd project

obfuscated:
* This is my custom keyboard project.
* I'm using mk1 for several years already. I've now finished mk2.
* I'm requesting a pid, becuase I'm doing some arduino development and it is
  annoying that my active keyboard clashes with the other arduino boards.
* There is a slight chance that mk2 would see a production run in small
  numbers after some testing done by me if the concept is working as
  expected.

### DIFF
--- a/1209/ADA1/index.md
+++ b/1209/ADA1/index.md
@@ -1,0 +1,9 @@
+---
+layout: pid
+title: thekbd
+owner: obfuscated
+license: GPL3
+site: https://github.com/obfuscated/thekbd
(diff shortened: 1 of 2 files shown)


Arachnid:
Thanks for your PR!

If you only need a PID to use for development purposes, feel free to use the reserved development PID, 0x1209/0x0001. Please feel free to open a new PR at such point as you need a PID for external use.

PR #391, opened 2018-12-27
Add PID request for Cherry Pi 11411's Optical Flow Board

cherrypi11411:
We are requesting a PID for our Optical Flow Board project


Arachnid:
Nice project! I hope it meets with success.

PR #392, opened 2018-12-27
SAMboyantino PID Request

SandeepanSengupta:
Requesting new sets of PIDs (**E0E0**, **E0F0**, **EEEE** and **EEEF**) for a firmware/bootloader (a **CDC-MSC composite** bootloader that **requires two sets of ID** for **each emulation**, one in **bootloader** mode, one when running **firmware** code)

### DIFF
--- a/1209/E0E0/index.md
+++ b/1209/E0E0/index.md
@@ -1,0 +1,8 @@
+---
+layout: pid
+title: SAMBoyantino CDC Firmware
+owner: Sandeepan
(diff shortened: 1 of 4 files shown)


SandeepanSengupta:
I am requesting a different set of PIDs to avoid conflict as the firmware is implemented in a different hardware

Arachnid:
That's fine, but you still haven't provided any firmware in your repository.

PR #408, opened 2019-02-04
Replace PID ABB9 with PID 50AC. Requesting PID 50AD as well

eftaxiopoulos:
You had previously provided us the PID ABBA for our device called SafeWISE. Some time ago I requested a new PID (ABB9) because in the latest firmware upgrade, the bootloader will have a different PID than the main firmware and you provided this PID as well.

I thought that I could keep using the PID ABBA for the main firmware. However I realized that for the new firmware, both PIDs of the bootloader and firmware have to be different than the PID of the previous firmware (ABBA).

Therefore, if possible, I would like to release the PID ABB9 and replace it with the PID 50AC for the bootloader. And for the main firmware of the upgraded version I would like to request the PID 50AD.


Arachnid:
Can you explain why you can't reuse the old PID for either purpose? We would not be able to release the PID for reuse, so it would be permanently burned.

eftaxiopoulos:
I did not know that you could not release a PID and it would be burned. The reason I wanted to do this is because I wanted the new VID and PID to be very different that the previous PID (ABBA). That's why I asked for the 50AC and 50AD.

But of course there is no reason to burn a PID, so I could keep the PID ABBA and ABB9 and ask the PID ABB8. Would this be ok? If so, I will make a new pull request.

I would like to know too if I could change at least the product description on the PID ABB9 on my new pull request or if I should not modify anything on the index.md.

Arachnid:
We can't reuse a PID, because there may be hardware out there that's using that PID.

What do you need PID ABB8 for? It's still not clear  to me why you can't reassign this existing PID.

You're welcome to submit PRs amending the descriptions of your existing PIDs.

eftaxiopoulos:
Our device SafeWISE, as I have noted in the index.md, uses the TREZOR One firmware (https://github.com/trezor/trezor-mcu).

So the first version of TREZOR One used HID communication and had a single PID for both bootloader and firmware. This is why we requested the PID ABBA months ago.

Later, TREZOR made a big firmware update changing the communication to WebUSB. As you can see in the code, they used 2 new PIDs for the new bootloader and firmware (that they acquired from you). So we need also 2 new PIDs.

In the beginning I thought that I could ask only 1 new PID for the bootloader (ABB9) and use the previous PID (ABBA) for the new firmware. And this is why I requested only the PID ABB9. But as I realized now, after testing the new firmware, it needs to have a different PID than the old firmware because the code decides if it is the HID or WebUSB version based on the firmware PID.

So finally it should be like this:

HID version : PID ABBA for both firmware and bootloader
WebUSB version: PID ABB8 for the bootloader and ABB9 for the firmware

Note: I wanted to know if I can make changes on the title of the PID that you have provided me already. I would like to modify PID ABBA from "SafeWISE" to "SafeWISE (HID)" and PID ABB9 from "SafeWISE Bootloader" to "SafeWISE". Please let me know if a renaming is possible so that I can prepare my next PR.

Arachnid:
Okay; in that case, please request two new PIDs, but leave the existing one in place. You're welcome to submit PRs updating details of existing PID allocations.

PR #410, opened 2019-02-12
Request PID for Hideez Wallet

dzaliznyak:


dzaliznyak:
Hardware source files have been added

PR #411, opened 2019-02-14
Modify description in PIDs ABBA and ABB9 and request PID ABB8

eftaxiopoulos:
Following our conversation in my previous PR, I request one new PID (ABB8) and I made some modifications on the description of the 2 PIDs that you have already provided us.

### DIFF
--- a/1209/ABBA/index.md
+++ b/1209/ABBA/index.md
@@ -1,6 +1,6 @@
 ---
 layout: pid
-title: SafeWISE
+title: SafeWISE (HID)
 owner: CoinWISE
 license: GNU Affero General Public License v3.0
 site: https://safewise.io/
--- a/1209/ABB8/index.md
+++ b/1209/ABB8/index.md
@@ -1,6 +1,6 @@
 ---
 layout: pid
-title: SafeWISE
+title: SafeWISE Bootloader
 owner: CoinWISE
 license: GNU Affero General Public License v3.0
 site: https://safewise.io/


eftaxiopoulos:
I don't really get the travis error

peternewman:
Hi @eftaxiopoulos ,

I think the Travis error may be a red herring, @Arachnid can probably get it re-run which will hopefully chase it away, or you could add another commit.

eftaxiopoulos:
Thank you @peternewman 
I saw that in the PR submitted after mine, there is the same travis error. I think that another commit will not solve the problem. @Arachnid what is your opinion?

peternewman:
Oops, yes well spotted. Anyway, the error is something @Arachnid will need to fix, rather than anything up with your changes:
https://travis-ci.org/pidcodes/pidcodes.github.com/jobs/494756816#L3115-L3117

```
bundle exec github-pages health-check
Checking domain pid.codes...
Uh oh. Looks like something's fishy: The custom domain for your GitHub Pages site is pointed at an outdated IP address. You must update your site's DNS records if you'd like it to be available via your custom domain. For more information, see https://help.github.com/articles/setting-up-a-custom-domain-with-github-pages/. (DeprecatedIPError)
```

1209/ABBA/index.md
```diff
@@ -1,6 +1,6 @@
 ---
 layout: pid
-title: SafeWISE
+title: SafeWISE (HID)
```
Arachnid: Do you need a separate PID for HID? Can't you expose both HID and other profiles on the same PID?

eftaxiopoulos: Yes I need a diferent one. I will copy here the explanation from the previous PR that I submitted:

"Our device SafeWISE, as I have noted in the index.md, uses the TREZOR One firmware (https://github.com/trezor/trezor-mcu).

So the first version of TREZOR One used HID communication and had a single PID for both bootloader and firmware. This is why we requested the PID ABBA months ago.

Later, TREZOR made a big firmware update changing the communication to WebUSB. As you can see in the code, they used 2 new PIDs for the new bootloader and firmware (that they acquired from you). So we need also 2 new PIDs.

In the beginning I thought that I could ask only 1 new PID for the bootloader (ABB9) and use the previous PID (ABBA) for the new firmware. And this is why I requested only the PID ABB9. But as I realized now, after testing the new firmware, it needs to have a different PID than the old firmware because the code decides if it is the HID or WebUSB version based on the firmware PID.

So finally it should be like this:

HID version : PID ABBA for both firmware and bootloader
WebUSB version: PID ABB8 for the bootloader and ABB9 for the firmware"

PR #420, opened 2019-04-20
Request PID for dme - an ergonomic keyboard

hidsh:
Hello pidcodes,

I would like to get PID for my second USB keyboard which is start to develop.

Thank you!


hidsh:
I will retry to request later. 
Thank you.

PR #421, opened 2019-04-26
Added Robotics-Masters and the MM1 Hat

hmic:
We support a set of different bootloaders and firmware choices for our board to make it as accessible as possible for the different communities, for this reason we request differnet VID for each of the boards usecases.
We are working on adding Ardupilot and Dronecode support at the moment and would like to request continuous VIDs 0x4D46 and 0x4D47 for the programming USB device in those i the near future.

### DIFF
--- a/1209/4D40/index.md
+++ b/1209/4D40/index.md
@@ -1,0 +1,9 @@
+---
+layout: pid
+title: MM1 Hat - Base
+owner: Robotics-Masters
+license: OHL
+site: https://robotics-masters.co/
+source: https://github.com/robotics-masters/mm1-hat-schematics/
+---
+MM1 Hat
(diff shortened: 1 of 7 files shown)


hmic:
Can you please update/run travis again? I fixed the error reported and force-pushed to the fork not to have many commits in this. But travis does not recognise the change and retriggers.
If not possible, I can still push a new commit to trigger travis again...

hmic:
Hi Nick,
sorry for messing things up.

I fixed the changed repo URL, renamed the LICENCE to LICENSE in the schematics repo and have Cian add a LICENSE file to the bootloader repo.

Regarding the board VID, it will be used in production with an automated testing device, but we will flash the bootloader before shipping, so it will not be seen in the wild.

Arachnid:
If the PID for the board won't be used in the wild, could you use one of our test PIDs, instead of allocating a public one for it?

PR #450, opened 2019-08-15
Request for 0x1209/0xBEE5 VID/PID pair

Duckle29:
Used for an open-source keyboard


Duckle29:
The hardware sources are all MIT licensed. The firmware that it runs is GPL-2.0 I don't know if I can list multiple or not

Duckle29:
As this is a USB-HID device I moved to the 0x0--- range

Duckle29:
I noticed there were no orgs with spaces, so I looked through a few old PRs and found this: https://github.com/pidcodes/pidcodes.github.com/pull/301 as such I changed my org to be without spaces.

Duckle29:
Okay, I misinterpreted what was meant by that. Back to my original request of 0xBEE5 :)

PR #452, opened 2019-08-20
Add org entry for DorianTech

dorian-tech:
I'm adding an organization entry for DorianTech.


Arachnid:
Thanks for your PR! Do you want a PID? If so, you can submit the org entry in the same PR as the request for the PID.

PR #457, opened 2019-09-03
Reserve 0x214C, 0x214D for passinglink.

jmgao:


parkrrrr:
In case you're wondering why the checks failed, it looks like your org request file has only two -- for the dividing lines instead of the required three. These things are read by machines, and the machines are picky.

jmgao:
Oops, thanks for the tip! Fixed and rebased.

PR #461, opened 2019-09-17
Request to add Etherkit Empyrean Alpha/Beta

NT7S:
Requesting to add Etherkit as an organization and the devices Empyrean Alpha (0x1209/0x73ea and 0x1209/0x88ea) and Empyrean Beta (0x1209/0x73eb and 0x1209/0x88eb)

### DIFF
--- a/1209/73EA/index.md
+++ b/1209/73EA/index.md
@@ -1,0 +1,8 @@
+---
+layout: pid
+title: Empyrean Alpha
+owner: Etherkit
+license: GPL3
+site: http://www.etherkit.com/
+source: https://github.com/etherkit/Empyrean/
+---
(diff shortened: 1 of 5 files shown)


Arachnid:
Why do you need two PIDs for each device?

NT7S:
The bootloader firmware (this is an Arduino IDE compatible microcontroller) enumerates in two different ways. In normal usage it enumerates as a USBCDC (virtual serial) device. However, when reset is double-tapped or the Arduino IDE initiates a sketch upload, it enumerates in sketch upload mode as an HID device. This is a derivative of the Arduino Zero, which works in the same way. Thanks

Arachnid:
Please specify in your PID files which ones are for bootloader mode, then.

NT7S:
Should that be specified in the title or somewhere else?

PR #466, opened 2019-09-28
Request PID 0x6060 for keyboard

rpedde:
QMK based firmware in /qmk, hardware in kicad in /schematic
with gerbers linked from the README.

All under GPL2 (or later)

### DIFF
--- a/1209/6060/index.md
+++ b/1209/6060/index.md
@@ -1,0 +1,8 @@
+---
+layout: pid
+title: r68 65% keyboard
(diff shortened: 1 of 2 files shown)


rpedde:
It's not entirely clear to me why ci tests failed.  One appears to be gem version issues (maybe?) the other travis connectivity.  If the errors are indeed mine, I'm not sure what they are -- I'd appreciate guidance on how to resolve them.

Thanks.

PR #479, opened 2019-11-27
PID request for Parkrrrr/BDIOHID

parkrrrr:
[BDIOHID](https://github.com/parkrrrr/bdio) is a software-only project, because it is intended to run on any Linux SBC, such as the OrangePi Zero+ or the NanoPi NEO, that can run OpenWRT and has both USB host and USB device ports. 

When running, it translates proprietary communications protocols used by legacy Braille displays into [the recently-standardized HID Braille protocol](https://www.usb.org/sites/default/files/hutrr78_-_creation_of_a_braille_display_usage_page_0.pdf). This standard provides a number of benefits to the visually-impaired community, not least of which is a significantly lowered barrier to entry for software developers wishing to provide Braille input and output. I consider BDIOHID to be an integral part of that step forward, in that it endeavors to provide a path forward for users who have older hardware.

We cannot simply replicate the VID and PID from the original Braille hardware, because in many cases that VID and PID are tied to existing host drivers. For example, the HIMS SyncBraille, the first display supported by this project, is based on an FTDI FT232 USB-to-serial chip, so it is inextricably tied to the FTDI serial driver. So, even though this device exists only as a collection of bits on commodity hardware, it needs its own VID/PID.

### DIFF
--- a/1209/0002/index.md
+++ b/1209/0002/index.md
@@ -1,0 +1,7 @@
+---
+layout: pid
+title: Test PID
+owner: pidcodes
+license: any
+---
+This PID is reserved for use in private testing. Anyone may assign it to their device while they're testing in-house, but it MUST NOT be used on any device that will be redistributed, sold, or manufactured. Source code and configuration that references this VID/PID should warn users that the PID is not universally unique and should not be used outside test environments.
--- a/1209/0003/index.md
+++ b/1209/0003/index.md
@@ -1,0 +1,7 @@
+---
+layout: pid
+title: Test PID
+owner: pidcodes
+license: any
+---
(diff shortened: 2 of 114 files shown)


parkrrrr:
I know the commits look weird here. It's because I reused the branch where I'd created the 0x02-0x10 testing codes. That commit and the merge commit should cancel each other out.

chrysn:
In the spirit of peer-reviewing the PRs so when The Time Of The Next Merge comes all can be processed swiftly, I'd suggest you rebase your branch onto current master and force-push. That'll resolve the merge situation, and also give you an opportunity for more meaningful commit messages.

(Feel free to skip the rest of this text if you know how to do that anyway).

This is easily done when you have your branch checked out by running

    git fetch
    git rebase -i origin/master

removing the "Added additional testing codes" line (it's not applicable any more) changing the "pick" before the remaining "Create index.md" lines to "reword", opening up your editor for more descriptive commit messages. A `git push -f` will update both your branch on github and this pull request.

parkrrrr:
That didn't work, for reasons I won't go into. But I'll just do it the right way and abandon this PR.

PR #491, opened 2019-12-25
request PID 5741 for ArduPilot

tridge:
This requests a 2nd PID for ArduPilot, to allow us to support both single endpoint and dual-endpoint CDCACM devices on windows


tridge:
thanks!

PR #526, opened 2020-06-12
Added own organisation and PID

richardklingler:
First published USB project on Github, a simple 2-channel thermocouple sensor logger.
A new  version will be published soon based on existing design, and will use the same PID regardless if it is a 2-channel or 8-channel device.


tannewt:
Merged in #554

PR #528, opened 2020-06-23
Added lophtware's USB2.0 Type-C / PIC32 Breakout Board.

lophtware:
Hello,

This is to request a VID / PID of 0x1209 / 0xcb0b for the firmware that ships with my USB breadboarding widget.  The various sites are still being thrown together but the GitHub repository has all of the code, schematics, etc. so should meet the requirements set out in http://pid.codes/howto/

Any problems or questions, let me know.

Thank you.


tannewt:
Merged in #554

PR #535, opened 2020-07-15
Use same PID code for ODrive v4 also

madcowswe:
We are soon coming out with a new board revision with compatible interface. Updating the description to be more general.
Thanks.


tannewt:
Merged in #554

PR #539, opened 2020-07-26
Submit PID request for Abies

fsodari:


tannewt (changes requested):
Where is the source for the software side? Thanks!

fsodari:
Thank you for your response. I actually just finished up a project two days ago and master branch is stable and working exceptionally well!

I have to apologize. I was hoping I would be further along by now, and I actually had to postpone the project that I requested this pid for. I instead chose to make a smaller prototype to test the USB functionality. I have since completed this project and both the hardware and firmware are here. I'm in process of switching everything to MIT license from GPL if that's okay. Forever and always open source.
https://github.com/half-spin/headphone_dac_proto
https://github.com/half-spin/hp_dac_proto_fw

This project is an Asynchronous USB Headphone amplifier. It uses a PSoC 5LP microcontroller as a USB to I2S bridge. There is a Cirrus Logic DAC driving a TI headphone amplifier (Kicad project contains all part numbers). Objectively speaking, it sounds really good. I'm extremely happy with it. It could probably use some small improvements with the asynchronous feedback, but those will happen soon. We will also be testing limited DSP functionality with the available processing resources as well.

The main project "Abies" will use the exact same USB interface as the prototype. It will just feature an FPGA on the board as well. Can I use the same PID for this project as that? Abies won't be finished for quite awhile, but it will be open source hardware, firmware, and software too. I also have two additional, similar projects in the pipeline. They all will have the same functionality of "Asynchronous USB Audio Out/In" and will load the default windows driver. I'm happy to use the same PID for all of them if I can. Copy/Pasting USB code is my greatest skill. Just let me know what works! Thanks so much. I'm so happy this service is available. Let me know if I can contribute in any way.

tannewt (changes requested):
Hi! It's generally better to have unique PIDs for each product. That way it's easier for folks to identify what they have. So, I'd suggest updating this PR for the prototype and then getting a second PID for the FPGA board.

Please make sure the source URL points to either the software or hardware repo and that it links to the other repo.

fsodari:
In that case, I don't think this prototype project truly needs a PID. I will resubmit this PR when my main project exists. Thanks.

PR #544, opened 2020-08-08
Requesting PID 0D11 and 0D10 for Odiin

nitz:
Requesting the pids 0x0D11 (Odiin Application) and 0x0D10 (Odiin Bootloader) for the Odiin device (https://github.com/nitz/Odiin)


tannewt (changes requested):
Please switch to PIDs that don't start with 0 or 1.

nitz:
Ah, my bad. I assumed because I was using all 'generic' interfaces (CDC, MSC) that the 0x0000-0x0FFF range was okay. I've moved them to 0xD10F and 0xD110.

tannewt (approved):
Looks good!

PR #546, opened 2020-08-23
Updated usage of 4321

limpkin:
As we didn't end up using this PID for the mooltipass arduino bootloader, we are reusing it for the mooltipass mini ble


tannewt:
Merged in #554

PR #548, opened 2020-08-25
add rrainey organization and associated pid codes

rrainey:
Hi. I have created two new open hardware projects and I'm finding that I need distinct PID codes in order to interact with some development tools and drivers cleanly.


tannewt:
Merged in #554

PR #551, opened 2020-09-03
add LinklayerLabs org and CANtact Pro device (0x1209/0x1111)

ericevenchick:
Hello, I'd like a VID/PID for the new CANtact Pro device. This is an open source hardware / firmware CAN to USB tool. The VID/PID will be used for an updated driver which enables CAN-FD.

### DIFF
--- a/1209/1111/index.md
+++ b/1209/1111/index.md
@@ -1,0 +1,9 @@
+---
+layout: pid
+title: CANtact Pro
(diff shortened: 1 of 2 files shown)


tannewt (changes requested):
Please update the source link and switch to a PID that doesn't start with 0 or 1.

PR #554, opened 2020-09-15
Aggregate PR

tannewt:
This PR aggregates outstanding PRs into one so that pending PRs will not have overlapping PIDs. Hopefully it'll be merged soon.

Existing PRs will be merged into this PR if they are clearly and openly documented and do not break the site build. Others will be skipped with a note about why below.

Closed source projects are expected to purchase their own VID: https://www.usb.org/getting-vendor-id

Fixes #549 

Here is the status of outstanding PRs:
- [x] [PR 493 - Fixed index.md file(s)](https://github.com/pidcodes/pidcodes.github.com/pull/493) Not merged, author asked for PR to be closed.
- [x] [PR 494 - added AXIOM Beta and AXIOM Remote devices ](https://github.com/pidcodes/pidcodes.github.com/pull/494) Merged
- [x] [PR 504 - Requesting PID FFB0 for OpenFFBoard](https://github.com/pidcodes/pidcodes.github.com/pull/504) Merged
- [x] [PR 508 - Request for PID 5750 with VID 0x1209 for CUAV autopilot.](https://github.com/pidcodes/pidcodes.github.com/pull/508) **NOT** merged. No native source files for the hardware.
- [x] [PR 510 - Request pid for Comrade P1](https://github.com/pidcodes/pidcodes.github.com/pull/510) Merged. Open source software only. Existing hardware.
- [x] [PR 513 - Add 1209/B737 - LrLite](https://github.com/pidcodes/pidcodes.github.com/pull/513) **NOT** merged. No native source files for the hardware.
- [x] [PR 515 - Updated website details](https://github.com/pidcodes/pidcodes.github.com/pull/515) Merged. Info update.
- [x] [PR 516 - added 0x0h.com and ours first device](https://github.com/pidcodes/pidcodes.github.com/pull/516) Merged. Hardware is public domain on EasyEDA and software is MIT.
- [x] [PR 517 - Request PID for LNE projects](https://github.com/pidcodes/pidcodes.github.com/pull/517) **NOT** merged. No obvious hardware files.
- [x] [PR 518 - Added the MIDI2CV modules.](https://github.com/pidcodes/pidcodes.github.com/pull/518) Merged.
- [x] [PR 519 - Requesting PID 0xCACE](https://github.com/pidcodes/pidcodes.github.com/pull/519) Merged.
[...]

PR #558, opened 2020-09-23
Requests USB ID 1209:3845

eeble:
Requests the USB PID 3845 for a new macropad keyboard in current
development.  Hardware is built and firmware is being written using QMK.

Keyboard specific code to be added to QMK will be placed in the qmk directory in the myfirstmacropad repository


tannewt:
Merged in #554

PR #561, opened 2020-10-01
Changes to Css bg color & text color

kishankrs:
Changes in Css file to make website more appealing.


tannewt (changes requested):
Please don't spam CSS changes for Hacktoberfest.

PR #562, opened 2020-10-03
Request PID 0x1337 for Schwurbler

mommel:
adds org  Happy Shooting Community
adds 0x1337 for Schwurbler Midi Device


tannewt (changes requested):
Please pick a different PID. 0x1xxx is reserved by the original VID owner. See step 3: http://pid.codes/howto/

mommel:
Will open a corrected one

PR #564, opened 2020-10-10
Adding me and my Board

Targett363:
Trying to get the new Beta version of Circuit Python to run on modules on my board, I needed PIDs hope I've done this right I'm on my first day actually using GIT!


Targett363:
Will pull again with correct licence listed!

tannewt (changes requested):
Please post the native hardware design files along with a LICENSE file for them in the repos. We don't consider schematic images open because they cannot be modified to produce different hardware.

Targett363:
I think I got it now. The jpeg was just a placeholder, I hadn't pushed the KiCad files! Thanks for all your help.

Targett363:
I think I've managed to do it right this time.
Thanks for your time!

tannewt (changes requested):
Please switch to PIDs that don't start with 0 or 1. Repo looks good! I'd suggest renaming the license file to LICENSE because it's a standard filename.

tannewt (approved):
Looks good! Thanks!

PR #565, opened 2020-10-11
Request 1209/3320 for MJS2020 sensor board

matthijskooijman:
This board is still in development, but the latest prototype version is already published with full sources, the final board will be published in the same place. The same USB id will be used for the prototype and final versions.

### DIFF
--- a/1209/3320/index.md
+++ b/1209/3320/index.md
@@ -1,0 +1,12 @@
+---
+layout: pid
+title: MJS2020: Meetjestad sensor station 2020
+owner: Meetjestad
+license: CERN-OHL-W
+site: https://www.meetjestad.net
+source: https://github.com/meetjestad/mjs_pcb/tree/MJS2020
+---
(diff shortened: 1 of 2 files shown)


tannewt (changes requested):
Please pick a PID that doesn't start with 0. They are reserved: http://pid.codes/1209/

matthijskooijman:
W00ps. I checked carefully and then decided to change the PID just before submitting the PR, should have rechecked carefully.

I've rebased on top of master and changed the PID to a non-reserved value.

matthijskooijman:
I was also planning to submit a PID checker script with github integration, but just noticed that it was already added in the recent merging spree, great! I have a small clarification for the docs, though, which I'll submit in a separate PR.

tannewt (approved):
Looks good! Thanks!

matthijskooijman:
Thanks!

PR #567, opened 2020-10-20
Clarify reserved PIDs in the howto

matthijskooijman:
Previously, this page would refer to the VID page for the reserved PIDs,
but also list the 1xxx reservation explicitly while omitting the 0xxx
reservation. This commit lists both, but more briefly, removing some
duplication.


tannewt (approved):
Looks good! Thank you!

PR #577, opened 2020-10-30
Fix MJS2020 (1209/3320) board page

matthijskooijman:
This used a colon in the title, which breaks parsing of the frontmatter
block. This causes the board to be omitted from the pid list and the
board page to only show the extended description, none of the defined
fields.

Adding quotes around the title fixes this.


tannewt (approved):
Thanks for the fix!

PR #527, opened 2020-06-12
uCANDevices

Lukasz-Juranek:
Please and 3 new PIDs (0777,0776,0775) for awesome open-source products from https://ucandevices.github.io/

PR #581, opened 2020-11-17
Restore System76 Thelio Io mapping, destroyed by #527

jackpot51:
The original mapping was in #361


tannewt:
Sorry about that! Thanks for the fix!

tannewt (approved):

PR #582, opened 2020-11-18
Additional validation.

tannewt:
* Ensure PID directories only have an index.md
* Ensure PIDs are valid hex.

@joernalraun Your original PID CM20 isn't valid hex so I moved you to CA20.

@Lukasz-Juranek I've corrected your 2779. You should also check all of your PIDs are OK.


tannewt:
Merging since no one has responded.

PR #584, opened 2020-11-19
Add PID EA69 by pgum (gumulka)

pgum:
Requesting PID EA69 for open source midi controiller based on arduino leonardo.


tannewt (approved):
Looks good!

PR #594, opened 2020-12-06
Request PID 5C00

zgtk-guri:
Request 1209:5C00 for a Pro micro compatible mcu board with a USB Type-C receptacle.
https://github.com/zgtk-guri/c-pro-micro
VID/PID will be used with Catalina bootloader.


tannewt (approved):
Arduino compatible. Kicad files available.

PR #605, opened 2021-01-14
Requesting PID for Double M33 Express

gsteiert:
Requesting PID for my Double M33 Express board.
A link to the software source is included.
The hardware source will be released after the CrowdSupply campaign completes.


tannewt (changes requested):
Sorry, we don't give out pid.codes PIDs until the software and hardware files are available. Looks like you could get one from NXP though for free: https://community.nxp.com/t5/Kinetis-Microcontrollers/NXP-USB-VID-PID-Program/ta-p/1124867

PR #610, opened 2021-01-25
KoBussLLC PID BA00 and BA01 for Grabert and Squash Mechanical Keyboards

camrbuss:
Requesting BA00 and BA01 for two open-source mechanical keyboards. The hardware is open-source using KiCad and FreeCAD, and the firmware is enabled by the open-source project QMK.


tannewt (approved):
They look neat!

PR #611, opened 2021-01-27
Generation of pid request for Orbtrace

mubes:
Hi there,

We are working on an (Open) USB-based TRACE sniffer for CORTEX-M. It is under a BSD licence. The hardware is based on an ECP5 FPGA and USB-HS (eventually USB-SS, but that's step 2). It would be appreciated if we can get a PID for this device. Full bitware and supporting software is at the links provided in the request.

Thanks in advance

DAVE

### DIFF
--- a/1209/3443
+++ b/1209/3443
@@ -1,0 +1,8 @@
+---
+layout: pid
+title: Orbtrace
+owner: Orbcode
+license: BSD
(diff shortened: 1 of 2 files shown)


tannewt (approved):
BSD software only. Looks good!

PR #616, opened 2021-02-11
Register WindowMaster as 1209:4573

agausmann:


tannewt (approved):
Neat to see embedded Rust!

PR #618, opened 2021-02-15
RIOT: Allocate 7D02 for devices executing the shipped bootloader

chrysn:
This is an extension to the previously allocated pairs 1209/7D00 and /7D01.

As anticipated back then, there is now a USB bootloader, which it makes sense to distinguish from the regular applications. This PR is allocating 7D02 for that purpose.


tannewt (approved):
Thank you for taking a look at other PRs!

PR #619, opened 2021-02-16
1209/7D02: License correction

chrysn:
I made a mistake when building 7D02 in #618 copying from 7D01; this fixes the oversight.

("any" applies to general applications built on RIOT; the code only applies to the DFU bootloader shipped with the OS, which is LGPL).


tannewt (approved):
Thanks for correcting this!

PR #621, opened 2021-02-19
PID Request for CH552E development board.

ichirowo:


tannewt:
What software will be using this PID? It must be open source as well.

ichirowo:
I see, that's what you mean.
 I understand now.

PR #622, opened 2021-02-19
added org and PID for librelab::Big Dot board.

unixjazz:
request for PID for the librelab::Big Dot, SAML-21 based board.
** thank you ** for supporting oshw dev!


tannewt (approved):
Looks good! I'd recommend separate PIDs for the bootloader and arduino.

PR #627, opened 2021-03-14
Request for PID 0xFFCD for USB Serial Multiplexer

TMakins:


tannewt (changes requested):
Source link doesn't work. Have you pushed the files?

TMakins:
Sorry, accidentally published as private - should be resolved now. Thanks

tannewt (changes requested):
Please provide the original PCB schematic and layout files. PDFs aren't considered open for pid.codes because they aren't modifiable.

PR #628, opened 2021-03-16
add Kavoliunas organization with 69A1 PID

CyberGear:


tannewt (approved):
Software only on pro-micro with QMK.

PR #629, opened 2021-03-22
PID Hunter Cat NFC

sabas1080:
Hi

I want to request a PID for the HunterCatNFC

https://github.com/ElectronicCats/HunterCatNFC

Thanks


tannewt (approved):
Your README states the license but you probably want to add the `LICENSE` files it references too.

PR #633, opened 2021-04-06
Request Addition of self.

Envious-Data:
would like to utilize PID for a open source keyboard ive designed which has a type-c port


tannewt (approved):
Looks good for the VID. Please rebrand your custom CircuitPython if you don't plan on committing your changes into mainline. Only builds from the adafruit/circuitpython repo should be branded as CircuitPython.

PR #634, opened 2021-04-09
Add PID for ProjectStida RoboCloud

2231puppy:
Hi, I'm working on an MCU called the RoboCloud. It is an ESP32-S2 motor driver board with Adafruit Stemma QT/Sparkfun Qwicc and a bunch of exposed IO pins. It is designed to drive 2 4.5A motors. I'm building CircuitPython and need a PID. If that's ok, great! If not, tell me what I need to fix and I'll do it. Source files are at https://github.com/2231puppy/RoboCloud

Thanks!

### DIFF
--- a/1209/1234/index.md
+++ b/1209/1234/index.md
@@ -1,0 +1,7 @@
+---
+layout: pid
+title: RoboCloud
+owner: myorg
+license: MIT
+source: http://github.com/2231puppy/RoboCloud/
+---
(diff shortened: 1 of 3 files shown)


tannewt (changes requested):
What software are you running on your boards? It needs to be open source as well.

2231puppy:
I'm running Adafruit's CircutPython

2231puppy:
I actually just got a PID from Espressif, closing now.

PR #643, opened 2021-05-03
Add PID 0x4040 for PicoHX FPGA programmer

dan-rodrigues:
This PR adds the `drr` org and the PID mentioned in the title.

The PID is for [this PCB](https://github.com/dan-rodrigues/pico-hx) where an RP2040 implements a USB device that can program an iCE40 FPGA using a bitstream sent by the host.


tannewt (approved):
Looks good!

PR #646, opened 2021-05-06
Added the organisation rykkers and added request for 2 pid BBBB and BBB2

rykkers:
As above, org added and 2 pids added

### DIFF
--- a/1209/BBB2/index.md
+++ b/1209/BBB2/index.md
@@ -1,0 +1,10 @@
+---
+layout: pid
+title: bb w_less tact usb2
+owner: rykkers
+license: MIT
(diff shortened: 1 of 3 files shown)


tannewt (changes requested):
For the hardware to be considered open source, you need to provide the pcb layout file as well.

rykkers:
> For the hardware to be considered open source, you need to provide the pcb layout file as well.

Sorry, that's my bad, "brd" file wasn't added to the commit it's now included: https://github.com/rykkers/w_less_tact

tannewt (approved):
Thanks!

PR #649, opened 2021-05-19
Pull Request for LotusProLLC

LotusMech:
We are an organization requesting VID 1209 PID 7272 and possibly a few more in the future 
(namely 7273 727A 7276 7278 952B 2B95) as we release more products in the near future. If you can hold on to these for us as we continue to release and update our fork, please let us know.

### DIFF
--- a/1209/7272
+++ b/1209/7272
@@ -1,0 +1,8 @@
+---
+layout: pid
+title: Lotus 2k
+owner: LotusProLLC
+license: GNU General Public License v3.0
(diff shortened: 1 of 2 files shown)


tannewt (changes requested):
The hardware files must be provided and the source code must as well. All I see is trivial keymapping structs and hex files. We do not hold PIDs either. If you are a business, then I'd recommend purchasing your own VID.

PR #650, opened 2021-05-19
Added new org caltabid and new Device 1209/DACA for MidiPedalBoard

caltabid:
Added new org caltabid; Added new Device 1209/DACA for project MidiPedalBoard


tannewt (approved):
Looks good! Hand wired but documented.

PR #658, opened 2021-06-30
PID allocation request for bad_alloc's dragonBoot

DX-MON:
Prefered PID for dragonBoot is BADB

The project is licensed under BSD-3-Clause, as are all projects on which dragonBoot depends (dragonUSB, dragonTI, substrate).

License for dragonBoot: Software

### DIFF
--- a/1209/BADB/index.md
+++ b/1209/BADB/index.md
@@ -1,0 +1,21 @@
+---
+layout: pid
+title: dragonBoot
+owner: bad_alloc
+license: BSD-3-Clause
+site: https://github.com/bad-alloc-heavy-industries/dragonBoot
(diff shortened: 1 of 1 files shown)


tannewt (approved):
Looks good. I recommend TinyUF2 as a USB bootloader because it does mass storage: https://github.com/adafruit/tinyuf2

DX-MON:
appreciate the input - in this case, I've been able to fit DFU in 5.7kiB on the Tiva-C, and more importantly to my application, it provides additional routines that the application can call for handling things like application page programming on the ATXMega256A3U, allowing the firmware to safely rewrite itself (to use Flash as EEPROM).

PR #659, opened 2021-07-10
Added 1209/EDCA for the educator version of a Monero hardware device

michaesc:
We recently designed (not yet in production) a simpler version of a Monero hardware device, to be used by educators when teaching classes to those interested in getting started with cryptosecure electronics. To help define such devices (for example when connecting on a computer already connected to other devices) we request the new PID EDCA. Thank you!

### DIFF
--- a/1209/2E7F/index.md
+++ b/1209/2E7F/index.md
@@ -1,0 +1,11 @@
+---
+layout: pid
+title: Midikraken Bootloader
+owner: Windfisch
+license: Apache 2.0 and CERN OHL-S-2.0
+site: https://github.com/Windfisch/midikraken
+source: https://github.com/Windfisch/midikraken
+---
+
+Midikraken is an open-source, open-hardware MIDI-USB-interface supporting up to
+16 MIDI ports, with more features to come.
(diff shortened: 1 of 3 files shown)


michaesc:
> Where is the hardware source and what software does it run?
>
The CERN OHL hardware source is in the **hardware** subdirectory of the following link:

https://github.com/monero-project/kastelo/

This hardware is embedded and powered by a microcontroller, so there is no host software. The USB logic is implemented in firmware and is in an early state of development. The firmware source belongs in the **firmware** subdirectory of the same link.

We have a different firmware project (BSD licensed) that runs on the device as well, and it is here:

https://github.com/m2049r/monerujo-hw/

Does that answer your question?

michaesc:
I hope my answer to 'where is the hardware source' was sufficient, and that we're able to use the PID soon. Thanks!

tannewt:
The hardware directory was last updated in 2019 but you said in the first post:

> We recently designed (not yet in production) a simpler version of a Monero hardware device

I assume this PID is for the newer device. Until it is released, then please use a test PID.

PR #660, opened 2021-07-11
Pi-KVM organization and IDs

mdevaev:
Hi! Pi-KVM is a project to create an open source KVM-over-IP based on Raspberry Pi. All software is distributed under the GPLv3 license, the basic schematic technology is also open.

I'm asking you to allocate two identifiers for different USB emulation devices. They will be used to make it easier for system administrators to distinguish them from generic Arduino and other OTG devices.

https://github.com/pikvm/pikvm

### DIFF
--- a/1209/EDA1/index.md
+++ b/1209/EDA1/index.md
@@ -1,0 +1,11 @@
+---
+layout: pid
+title: Pi-KVM Composite Device
+owner: Pi-KVM
+license: GPLv3
+site: https://pikvm.org
+source: https://github.com/pikvm/pikvm
+---
+Open and cheap DIY IP-KVM based on Raspberry Pi.
+
+This device can combine HID, Mass Storage, Ethernet-over-USB, and so on.
--- a/1209/EDA2/index.md
+++ b/1209/EDA2/index.md
@@ -1,0 +1,11 @@
+---
+layout: pid
+title: Pi-KVM HID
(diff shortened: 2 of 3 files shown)


tannewt (changes requested):
Where are the hardware files for the V3 HAT?

mdevaev:
V3 hat is a commerical hardware which uses open source software. This is a separate project. I'm not asking for a PID for this, but specifically for software that people already use in their DIY KVM. Schemes v0 and v2 are open.

tannewt (approved):
Ok. Please do not use this PID for images meant for the closed, commercial V3 hardware. These PIDs are for open hardware projects only.

mdevaev:
@tannewt Thanks, but the last thing I want to do is break your rules :) I really can't use this even if the v3 firmware is completely open and the codes are part of it, and not flashed on hardware? Firmware v2 and v3 have no differences on the part of the software and differ only in a couple of lines in the configuration files.

tannewt:
The goal of pid.codes is to promote open hardware by granting PIDs. You should open source your hardware if you want to use a pid.codes PID on it. If you don't want to because of commercial reasons, then you should be willing to pay for a commercial PID. It's another cost of doing business. Most microcontroller manufacturers will sublicense you one (some for free.)

mdevaev:
Got it.

PR #661, opened 2021-07-14
Add Rookie Drive in device mode as 1209/3434

Konamiman:
[Rookie Drive](http://rookiedrive.com/en/) is a cartridge for MSX computers featuring a CH376 chip and an USB-A port. Originally the cartridge was designed to work as an USB host, but the CH376 can be configured as a device as well; so just for fun I wrote [software to use a Rookie Drive as a device](https://github.com/Konamiman/NestorDevice).

Although it's an amateur product, Rookie Drive is not open source hardware; but it's not rocket science either: it's just a CH376 chip mapped to a couple of Z80 ports, and anyone with MSX hardware design knowledge can create a clone... which is exactly what [the MSX-USB project](https://github.com/S0urceror/MSX-USB) does; this one, yes, it's open source hardware.

So I've linked the MSX-USB project in the PID description file, but I prefer to keep the name "Rookie Drive" since it's the original project and is more well-known in the MSX community.

Video showing a MSX computer using a Rookie Drive in device mode to act as an USB keyboard: https://twitter.com/konamiman/status/1412516715318087681

### DIFF
--- a/1209/3434/index.md
+++ b/1209/3434/index.md
@@ -1,0 +1,9 @@
+---
+layout: pid
+title: Rookie Drive in device mode
+owner: Konamiman
+license: MIT (software), GPL 2 (hardware)
+site: https://github.com/Konamiman/NestorDevice/
+source: https://github.com/S0urceror/MSX-USB/
+---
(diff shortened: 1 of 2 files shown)


tannewt (approved):
Thanks for making open firmware!

PR #666, opened 2021-08-01
Add org bonzi, and PID 6939

bonzi:


tannewt (changes requested):
The keebs.bonzi.sh links don't work. Where is the hardware source?

bonzi:
Sorry about that, Repo's are now updated to reflect the correct project, and the website has been enabled (Sorry about its brevity, the project is still in development)

tannewt:
What software does it run?

I'd suggest a non-GPL license for hardware since hardware is different than software. (I won't block this merge though if you want to stick with GPL.) More info here: https://www.oshwa.org/faq/#what-license-to-use and here: http://www.ladyada.net/library/openhardware/license.html

bonzi:
> What software does it run?
https://github.com/qmk/qmk_firmware/

> I'd suggest a non-GPL license for hardware since hardware is different than software. (I won't block this merge though if you want to stick with GPL.) More info here: https://www.oshwa.org/faq/#what-license-to-use and here: http://www.ladyada.net/library/openhardware/license.html

I chose to use the GPL soley out of the fact that QMK uses it, However I am very much happy to switch to an OSHWA certified licence (After all, the KiCad project files and etc will be on GitHub).

bonzi:
CC BY-SA 4.0 licence has been implemented. Everything should be kosher now.

tannewt (approved):

PR #673, opened 2021-08-27
BOSS-Keyboards USB VID PID request

donicrosby:
USB VID PID request for our Muse QMK powered keyboard


tannewt (changes requested):
I think you need the PID owner capitalization to match the org folder capitalization to pass tests.

PR #676, opened 2021-09-02
Add pid codes for the Keyboardio Model 100

obra:
Hi! We've got a new keyboard coming out and would love to reserve the two PID codes in this PR. Please let me know if you'd like anything in this fixed up.

Thanks!

### DIFF
--- a/1209/2306/index.md
+++ b/1209/2306/index.md
@@ -1,0 +1,8 @@
+---
+layout: pid
+title: Model 100 Bootloader
+owner: keyboardio
+license: GPLv2
+site: http://keyboard.io
+source: https://github.com/keyboardio/Kaleidoscope
+---
--- a/1209/2307/index.md
+++ b/1209/2307/index.md
@@ -1,0 +1,8 @@
+---
+layout: pid
+title: Model 100
+owner: keyboardio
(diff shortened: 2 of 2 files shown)


tannewt (commented):
Hi @obra, where are the hardware source files? I don't give folks the PIDs until hardware and software source is published. (The testing PIDs can be used until then.)

tannewt (changes requested):
See above

obra:
Is this a policy change? In the past (for both the Model 100 and the Atreus), open firmware was considered sufficient — open source release of hardware designs wasn’t a requirement.

tannewt:
I don't think so. This says both: https://pid.codes/howto/

I do allow for firmware-only in the case that the applicant is making third party firmware. Otherwise, it is a carrot for folks to open source their hardware as well.

obra:
My apologies. Having gone spelunking, at the time we got the PID for the Keyboardio Model 01, the prose was less clear, but still had the same intent.  Because we don't intend to publish hardware design files before the product ships, we can't reasonably request a pid.codes PID for the Model 100. It was probably getting on time to join USB-IF anyway ;)

PR #685, opened 2021-09-17
Request for production test PID for ORBTrace

mubes:
To complete the 'suite' of PIDs for ORBTrace we'd really appreciate it if we could get a Production Test PID as it's going to make logistics a lot easier!
Thanks in advance
DAVE

### DIFF
--- a/1209/3441/index.md
+++ b/1209/3441/index.md
@@ -1,0 +1,9 @@
+---
+layout: pid
+title: Orbtrace-prodtest
+owner: Orbcode
+license: BSD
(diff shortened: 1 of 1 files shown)


tannewt:
Why not just use a test pid from pid.codes?

mubes:
That's a good idea. I'll pull back the PR. Only argument against it would
be if this gateware ever escaped to the wild, but (a) it only even goes in
ram, not flash and (b) folks could see that it's not a valid ID anyway.

Regards

Dave

Regards

Dave

PR #688, opened 2021-09-24
Requesting PID A182 for the RP2040 Stamp

arturo182:
Hi,

I would like to request PID A182 for the RP2040 Stamp, the schematics are licensed under the CERN Open Hardware Licence and available here: https://github.com/solderparty/rp2040_stamp_hw. 
It is my understanding that according to the "How to get a PID code" document, the schematics are enough to request a PID.
All the other KiCad design files will be released under the same license after the board goes into production.

Thank you :)


tannewt:
Please use a test code until all of the files are available. I'm not allowing for schematics-only in accordance with the OSHWA definition of open hardware:

> The hardware’s source, the design from which it is made, is available in the preferred format for making modifications to it.

arturo182:
All the hardware design files are now available under the CERN Open Hardware Licence: https://github.com/solderparty/rp2040_stamp_hw

tannewt (approved):
Looks good.

PR #689, opened 2021-09-25
PID requests for 4key & 6key keyboards (0x3305, 0x3306)

softplus:
I've made 2 more keyboards that I'd like to have separate PIDs for. The keyboards run QMK, and I have links to all relevant files that should be needed to produce them (schematics, EasyEDA JSON exports, BOM files). They're all under Apache 2.0 license. 

I also refactored the documentation on my side for the "2key" keyboards, which are under PID 0x3304, so I've updated the link here too.

### DIFF
--- a/1209/3304/index.md
+++ b/1209/3304/index.md
@@ -1,11 +1,11 @@
 ---
 layout: pid
-title: 3-Key-Ecosystem 2key2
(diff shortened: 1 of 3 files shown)


tannewt (approved):

softplus:
Thank you!

PR #691, opened 2021-09-29
Add djvs/aek under 1209/9209

djvs:


1209/9209/index.md
```diff
@@ -0,0 +1,9 @@
+---
+layout: pid
+title: aek
+owner: myorg
```
arturo182, line 4: This has to match the folder in `org`, for you, it should be `djvs`.

djvs: Whoops sorry! Just pushed fix.

djvs: @arturo182  All set?

arturo182: Sorry, I'm not the maintainer. I just pointed out the issue so you can fix it before @tannewt can merge.

tannewt (approved):

PR #693, opened 2021-10-05
Submitting request for a PID/VID registration for BrainBoardz

salimwp:
Hi there,

We would like to submit a PR for PID BAC0.

Thanks!
Salim


tannewt:
What device is this for specifically? Please include exported versions of the EasyEDA data so that others can modify the files. PDF or images is not enough.

tannewt (changes requested):
More details needed

PR #694, opened 2021-10-05
ZigStar USB/LAN PID

mercenaruss:
Add ZigStar Organisation and Devices


tannewt (changes requested):
0x1xxx PIDs are reserved.

Please provide modifiable hardware source files as well. Not just PDF schematics.

PR #697, opened 2021-10-28
Add DatanoiseTV PIDs.

DatanoiseTV:
Add PID and org for DatanoiseTV and RP2040 based PicoProbe++


tannewt (changes requested):
What hardware is used with this? If it is your own, it must be open source.

PR #708, opened 2021-12-22
Assign PID DE1A to Delta Pico calculator

AaronC81:
I'm requesting a PID for my Delta Pico scientific calculator, as it's now able to mount itself as a mass storage device.

This is my first PID, so I've also created an org for myself.

Thanks!


tannewt (approved):
Looks good! Thank you!

PR #716, opened 2022-01-14
Request PID 7110 for PicoDitDah Morse Code Keyer

dg1psi:
Hi,

I have created a morse code keyer using the Raspberry Pi Pico to connect a morse code paddle to a computer.
As the device acts as a custom USB device with two interfaces I would like to get an PID.

All the code is open source (MIT).


tannewt (approved):
Looks good.

PR #717, opened 2022-01-17
PID Request: TinyGo — A Go Compiler for Small Places

ardnew:
Requesting a unique PID for TinyGo-supported devices, which often already have a VID/PID specific to the vendor's officially supported firmware. Since TinyGo completely replaces this firmware and has developed its own independent USB software stack, it would be best to not reuse the original vendor's well-earned VID/PID, and instead we use our own separate PID with ties to the open source community.

### DIFF
--- a/1209/9090/index.md
+++ b/1209/9090/index.md
@@ -1,0 +1,8 @@
+---
+layout: pid
+title: USB
+owner: TinyGo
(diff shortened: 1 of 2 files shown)


tannewt (approved):
Neat!

PR #725, opened 2022-02-28
Request PID 5757 for various hobby projects

NickGuyver:
I wasn't sure what to put for device name since I have a few planned. Right now I have an adapter but very soon there will be a controller. Further down the road will likely be other totally different types of hardware.

Thank you for providing this service.


tannewt (changes requested):
Please use test PIDs until you have a specific open source device. https://pid.codes/pids/

PR #728, opened 2022-04-03
Request PID 1209:C12C for PicoSlave

janoskut:
Hey, I'm requesting `1209:C12C` for the PicoSlave project: https://gitlab.com/janoskut/picoslave.

Thanks!

### DIFF
--- a/1209/C12C/index.md
+++ b/1209/C12C/index.md
@@ -1,0 +1,10 @@
+---
+layout: pid
+title: PicoSlave
+owner: janoskut
+license: MIT
+site: https://gitlab.com/janoskut/picoslave
+source: https://gitlab.com/janoskut/picoslave/-/tree/main
+---
+# PicoSlave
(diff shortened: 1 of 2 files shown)


tannewt (approved):
The source looks ok. It runs on non-open hardware by a third party.

I'd suggest renaming the project away from the term `slave` in favor of `target` to match the latest spec and avoid a loaded term. https://en.wikipedia.org/wiki/I%C2%B2C#Revisions

janoskut:
Thanks for your review and approval @tannewt !

> I'd suggest renaming the project away from the term slave in favor of target to match the latest spec and avoid a loaded term. https://en.wikipedia.org/wiki/I%C2%B2C#Revisions

This is amazing news to me, I wasn't aware of recent changes to the terminology of I2C! Unfortunately I'm I afraid that I can't change the project name anymore, as it has made it's name already where we use it.

Is that, or the non-open source hardware a project for getting the open PID?

PR #740, opened 2022-05-27
Request PID 1209:E7E0 for IDEO eyeo-flower board

derekolson:
Requesting PID number 0xE7E0 for a custom made SAMD21E17D arduino compatible board. This board will be used at an upcoming workshop at [Eyeo Festival](https://eyeofestival.com/speaker/yc-jenna-dave/). We're hoping to contribute to the arduino ecosystem by showcasing a successful workflow getting arduino to boot on the SAMD21D hardware (which we haven't found much open-source information about).

FYI board source code and hardware files are still being finalized and will be uploaded soon.

### DIFF
--- a/1209/E7E0/index.md
+++ b/1209/E7E0/index.md
@@ -1,0 +1,10 @@
+---
+layout: pid
+title: Eyeo Flower
+owner: ideo
+license: MIT
+site: https://github.com/ideo/eyeo-flower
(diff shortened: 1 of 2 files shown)


tannewt (changes requested):
Please don't request a PID before the software and hardware sources are public and open sourced licensed. Use a test PID until then. A company like IDEO may be better off purchasing their own USB VID instead.

PR #745, opened 2022-06-13
add Kosagi / Precursor PID

bunnie:
Hi! We'd like to register a PID for our Precursor device core. Thank you!


bunnie:
Sure thing, I've updated the pins.

PR #748, opened 2022-06-16
Create PID C680 for picoframe

dedChar:
Requesting PID 0xC680 for a digital picture frame that uses USB to configure and upload images to the device.


jepler (approved):
Thank you!

Issue #754, opened 2022-07-20
htmlproofer 4.2.0 | Error:  invalid option: --internal-domains=pid.codes


bwshockley:
It appears the options for the html proofer check are out of date for 4.2.0

This error happens when checking my PR #753 

"htmlproofer 4.2.0 | Error:  invalid option: --internal-domains=pid.codes" 

https://github.com/gjtorikian/html-proofer/compare/v4.1.0...v4.2.0

bwshockley:
Appears that option "--internal-domains" was removed here:
https://github.com/gjtorikian/html-proofer/commit/7b0d193cf0796d7e722da3f65d7a47b8ba826a24

tannewt:
Fixed by #761

PR #759, opened 2022-08-03
Request Pid CC01 for CC Practice

mokesura:


mokesura:
I get an error below.
.github/workflows/htmlproofer.yml / HTML Proofer (pull_request)

What should I do to solve the problem?

Is it related to the following issues #754 ?
https://github.com/pidcodes/pidcodes.github.com/issues/754

tannewt (approved):

PR #761, opened 2022-08-05
Don't use htmlproofer 4.0.0 or newer

tannewt:
It removes internal_domains.


tannewt:
Merging since it fixes the CI.

PR #763, opened 2022-08-24
Add files for the E-Fidget PID

2231puppy:
Hi! I'm designing a PCB that has a circular arrangement of vibration motors on the back, controlled by an RP2040. When turned on and off in sequence, it will provide a similar tactile sensation that using a fidget spinner would.

I know this is probably one of the stranger PID requests you've gotten, but hear me out 😉 I want the firmware to be open source and easy to change the behavior of, so I want it to be CircutPython. As I'm sure you know, CircuitPython needs a PID to build, and as I want this board in the official CP repo, I'd need one. It's worth mentioning that only a small portion of the users of this board will want to modify the code, but I want it to be an option for those users.

Thanks for your time.

### DIFF
--- a/1209/EF00/index.md
+++ b/1209/EF00/index.md
@@ -1,0 +1,9 @@
+---
+layout: pid
+title: E-Fidget
+owner: 2231puppy
+license: AGPL-3.0
+site: https://2231puppy.tech/e-fidget/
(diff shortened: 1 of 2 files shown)


jepler (approved):
Intriguing!

PR #767, opened 2022-09-01
Adds new board Challenger RP2040 SubGHz to the list.

PontusO:
As the title suggests this is a request to get a PID for our new board Challenger RP2040 SubGHz. This board is supported both by the arduino-pico framework as well as circuitpython.


jepler (approved):
Thank you.

PR #770, opened 2022-09-08
Check frontmatter owners

jepler:
I noticed that not much validation of the metadata is performed. I identified that it would be useful to ensure that the owner's markdown file exists as expected. Other checks, such as for all required keys having values, could be added in the future.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -9,5 +9,5 @@ jobs:
       - uses: actions/setup-python@v2
         with:
           python-version: '3.x'
-      - run: python3 -m test.validate_pids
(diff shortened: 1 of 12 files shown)


jepler:
@tannewt ping!

peternewman:
> I noticed that not much validation of the metadata is performed. I identified that it would be useful to ensure that the owner's markdown file exists as expected.

I think that (at least used to) be caught by the link checking stuff, althought it seems that was broken, now fixed in #791 but the jekyll build warnings don't get flagged up with test failures which is a bit unfortunate.

jepler:
If this is redundant then please feel free to close it.

tannewt (approved):
Looks good to me! Thank you!

PR #779, opened 2022-10-12
add Pizza Robotics org & uNav2 pid

AlessioMorale:
Add Pizza Robotics org and VID for uNav2 robot motor controller


jepler (approved):
Thank you.

PR #780, opened 2022-10-15
Add Indices for Friend Team Inc and JukeBox

NotQuiteApex:


jepler (approved):
Thanks!

While checking the URLs I noticed that the FriendTeamInc URL comes up with a security warning, because the certificate is not valid for that domain. Mentioning it in case you were unaware.

jepler:
Oh, and there's a nit in the markdown code too:
```
  *  internally linking to /org/Friend Team Inc/, which does not exist (line 78)
     <a href="/org/Friend Team Inc/"></a>
```
The name needs to match the filename (without .org) exactly, so I think `/org/FriendTeamInc` will be right. Once you've corrected it so that the CI is 'green' :heavy_check_mark: please feel free to ping me

NotQuiteApex:
I noticed that security warning after the fact, it *should* be corrected now, thanks! I'll get the fixes done for Markdown now.

NotQuiteApex:
@jepler Alrighty, all fixed!

jepler (approved):
Thanks!

PR #781, opened 2022-10-20
Add pid for 42. Keebs Frood (mainly for CircuitPython)

piit79:
Another project in need of a PID :) The Frood is a Pro Micro replacement based on RP2040 with a standard pinout compatible with the SparkFun Pro Micro RP2040. Thanks!


jepler (approved):
thanks

PR #782, opened 2022-10-22
add PassPico PID

brxken128:
This adds `0xE128` for the [PassPico](https://github.com/brxken128/passpico) and the `brxken128` org.


jepler (approved):
Neat project, thanks!

brxken128:
> Neat project, thanks!

Thank you so much! It's still under heavy development but I'm pretty excited for it to become fully stable.

PR #784, opened 2022-10-30
Requesting PID for MultiBus Bridge

mringwal:
https://github.com/mringwal/multibus

MultiBus project aims to use common embedded interface such as 
GPIO, I2C, I2S, SPIO, SPIO, .. from a regular host system via
common development kits like Raspberry Pi Pico or ESP32.

First device with USB would be Raspberry Pi Pico.

Thanks!

### DIFF
--- a/1209/2007/index.md
+++ b/1209/2007/index.md
@@ -1,0 +1,12 @@
+---
+layout: pid
+title: MultiBus Bridge
+owner: MultiBus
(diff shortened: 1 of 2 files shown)


jepler (approved):
Thanks.

mringwal:
Thank you for providing this service!

PR #787, opened 2022-11-18
Add Custom 16MB Bluefruit Sense Build

DJDevon3:
Adafruit requires custom PID/VID for UF2 builds. Need PID for a 16MB flash version of the bluefruit sense.


jepler:
@DJDevon3 is this still needed?

DJDevon3:
No. Scott said I don't need one for a personal modification. Closing.

PR #789, opened 2022-11-27
Update name of OLP's RP2040 based device

peternewman:
Since @kripton changed it...


kripton:
Thanks, totally forgot to update this entry!

tannewt (approved):
Thanks!

PR #791, opened 2022-11-27
Try and fix the HTML proofer

peternewman:


peternewman:
Could you make the HTML Proofer a required check before merging is allowed now? Org owners can always override it if they need to.

PR #792, opened 2022-11-27
Add Uthol org and stakav2 PID

Uthol:


tannewt (changes requested):
Please post the pcb source files or schematic to use with a dev board.

PR #797, opened 2022-12-15
add shahnetwork org and pid

sagepeppermint:


tannewt (changes requested):
Where is the hardware documented?

sagepeppermint:
https://github.com/sagepeppermint/TBK-Mini

tannewt (changes requested):
Please include that link in the markdown files. Then you can link to QMK from there too.

PR #800, opened 2022-12-20
Add CH569 USB3 microcontroller board PID files

hansfbaier:
I just stress tested and verified the latest board revision. Works!
Ready for PID.
Signed-off-by: Hans Baier <hansfbaier@gmail.com>

### DIFF
--- a/1209/569C/index.md
+++ b/1209/569C/index.md
@@ -1,0 +1,9 @@
+---
+layout: pid
+title: USB3 Super Speed microcontroller board
+owner: OpenAudioGear
+license: CERN-OHL-P-2.0
+site: https://github.com/hansfbaier/ch569-usb3-board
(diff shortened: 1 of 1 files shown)


tannewt (changes requested):
Where is the bootloader source code? Hardware source looks good.

hansfbaier:
> Where is the bootloader source code? Hardware source looks good.

There is no bootloader source code yet. This is intended for the user of the board to use for his own bootloader,
because this is a generic development board.

tannewt:
Why do you need a PID if there isn't any code that will use it?

hansfbaier:
> Why do you need a PID if there isn't any code that will use it?

OK, I removed the bootloader from the PR.
Will make a separate one, when we would have one.

tannewt (approved):
Ok, sounds good.

PR #803, opened 2023-01-07
pid request for Let's Swamp

alg0002:


tannewt (changes requested):
What software will use this PID? I saw links to some supported software but not the source specific to this board. Thanks!

alg0002:
Thanks for the quick review comments.
I will use PID and VID with QMK Firmware. I uploaded the firmware to my repository.
https://github.com/alg0002/LetsSwamp/tree/main/firmware/QMK%20Firmware/lets_swamp

tannewt (approved):

PR #813, opened 2023-01-31
Looks like the URL for GitHub is Causing Issues

BrainBoardz:
I have updated the url for GitHub.


org/brainboardz/index.md
```diff
@@ -0,0 +1,7 @@
+---
```
tannewt, line 1: I think it'll be fixed once this copy of the org page is deleted.

tannewt (changes requested):
Please fix the CI. Software and firmware looks ok.

PR #814, opened 2023-02-01
Add libmcu MADI development board

onkwon:
Two endpoints are used, one for DFU and the other for CDC.

Schematics can be found [here](https://github.com/libmcu/development-board).
Source code can be found [here](https://github.com/libmcu/board-sdk).

Thanks.


onkwon:
Please let me PR later since hardware file is not in my hand.

PR #801, opened 2023-01-01
Added tinyVision.ai Inc. org, 0xB1C0 PID

vr2045:

PR #819, opened 2023-02-09
Added tinyVision.ai Inc. org, 0xB1C0 PID

josuah:
This is a continuation of #801 which unfortunately stalled without answer from our side.
Here is another try, rebased on latest `master`, with the original commit fixed a bit.

Thank you a lot for running this project.
A lot of headaches avoided to many people at once!


tannewt (approved):
Thanks for the follow up.

PR #821, opened 2023-02-13
EdgePro PID request

jturnsek:


tannewt (changes requested):
Please link to the open source software this runs and post modifiable ecad files. We don't consider PDF schematics open source.

PR #823, opened 2023-02-20
Add org: OneKeyLtd and pid:4F4A,4F4B,4F4C

424778940z:
We requesting more than one PID because they are used in different modes of the device.

### DIFF
--- a/1209/4F4A/index.md
+++ b/1209/4F4A/index.md
@@ -1,0 +1,9 @@
+---
+layout: pid
+title: OneKey Bootloader
+owner: OneKeyLtd
+license: Mix of licenses (LGPLv3, MIT)
+site: https://onekey.so/
+source: https://github.com/OneKeyHQ/firmware/tree/bixin_dev/docs/pcb
+---
+Onekey - Crypto wallet trusted by millions.
--- a/1209/4F4B/index.md
+++ b/1209/4F4B/index.md
@@ -1,0 +1,9 @@
+---
+layout: pid
+title: OneKey Device
+owner: OneKeyLtd
+license: Mix of licenses (LGPLv3, MIT)
+site: https://onekey.so/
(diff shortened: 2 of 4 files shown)


tannewt (changes requested):
Please link to your open source hardware repo and have its README link to the software repo.

424778940z:
> Please link to your open source hardware repo and have its README link to the software repo.

Sorry, my bad for not explain things in more details

Our hardware design and firmware is in the same repo, there is no need to link to other repo for hardware
Firmware: 
https://github.com/OneKeyHQ/firmware/tree/bixin_dev/
Hardware:
https://github.com/OneKeyHQ/firmware/tree/bixin_dev/docs/pcb

Our future device will also use the same PID and protocol, and it will have it's own repository. We also planning cleanning up current `firmware` repository and rename it to something make more sense. In the end, different devices will have it's own repository contains it's hardware desing and firmware source. Thus, we consider it is more proper to put our `OneKeyHQ `instead of the current `firmware` repo for the url part, and then people could easily find things they are looking for

lf you have other opinion / suggestings I'm also happy to hear them

tannewt:
Please link to the pcb directory from each pid page here. Otherwise it is hard to find. Please also add an explicit license to the pcb/ folder. It isn't clear what you intend because it is in the monorepo and LICENSE.md doesn't call out this folder.

424778940z:
> Please link to the pcb directory from each pid page here. Otherwise it is hard to find. Please also add an explicit license to the pcb/ folder. It isn't clear what you intend because it is in the monorepo and LICENSE.md doesn't call out this folder.

URL changed as requested.

GPL v3 license added to PCB folder as requested.
https://github.com/OneKeyHQ/firmware/tree/bixin_dev/docs/pcb/LICENSE.txt

PR #829, opened 2023-03-17
Add 01Space board for Circuitpython

Neradoc:
This is a PID request for running Circuitpython on the 01Space RP2040 0.42 LCD board.
https://github.com/01Space/RP2040-0.42LCD

The board is not open hardware, the maker has no VID and is not requesting a PID from Rapsberry Pi.

### DIFF
--- a/1209/42B0/index.md
+++ b/1209/42B0/index.md
@@ -1,0 +1,9 @@
+---
+layout: pid
+title: RP2040 0.42 LCD
+owner: 01space
+license: MIT
+site: https://github.com/01Space/RP2040-0.42LCD
+source: https://github.com/adafruit/circuitpython
+---
+Raspberry Pi RP2040 development board with a tiny screen.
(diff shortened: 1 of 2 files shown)


1209/42B0/index.md
```diff
@@ -0,0 +1,9 @@
+---
+layout: pid
+title: RP2040 0.42 LCD
+owner: 01space
```
tannewt, line 4: Please have yourself or "CircuitPython" as the owner so that it is clear that this PID wasn't granted to the manufacturer.

tannewt (changes requested):
Please use a different org.

1209/42B0/index.md
Neradoc: Ah, so should I not create the org then either ?

tannewt: Make one for yourself or CP but not the manufacturer.

Neradoc:
OK I'll think about it.

PR #830, opened 2023-03-18
Request PIDs for LibreVNA and LibreCAL

jankae:
Requesting one PID each for the two projects. Happy to split this into two PRs if you prefer it that way.


tannewt (approved):
All good. Repos are clear.

PR #841, opened 2023-04-29
add 6262 - xwhatsit QMK VIA

matthew-wolf-n4mtt:
Need a PID because the newest version of VIA does not like the 0xFEED VID.


tannewt (approved):
Open software. Someone else's hardware. (open-ish)

PR #851, opened 2023-06-14
PID request for Code60

mattyintegrals:
I created open source keyboard pcb. Pcb support open surce software QMK and VIA. Unique PID is required to recognize pcb by VIA.
My repository contains schematic and all files required to produce pcb like gerbers, bom and cpl.

### DIFF
--- a/1209/2121/index.md
+++ b/1209/2121/index.md
@@ -1,0 +1,10 @@
+---
+layout: pid
+title: Code60 Keyboard
+owner: mattyintegrals
+license: MIT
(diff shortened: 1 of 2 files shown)


tannewt (changes requested):
Please fix the repo link. It doesn't work.

mattyintegrals:
fixed

tannewt (changes requested):
That repo only has gerbers and pdfs. Please include the original eCAD files so that they can be modified. That's required to be considered open source.

PR #852, opened 2023-06-19
Add org and pid for sxkbd split keyboard (1209/DEF0)

sinitax:
https://github.com/Sinitax/sxkbd


sinitax:
*fixed link

sinitax:
For clarification: this is a firmware for crkbd-based split keyboard. The hardware is also open-source, available at: https://github.com/foostan/crkbd

PR #860, opened 2023-06-22
Request a VID/PID pair for zynthian USB gadget

jofemodo:


tannewt:
How does this use the USB PID? It looks like a big project that is mostly separate from USB.

PR #863, opened 2023-07-05
PID Request for Tiny Tapeout (0x5454)

urish:


tannewt (changes requested):
Where are the board design files?

urish:
There are preliminary versions of the PCB are here: https://github.com/TinyTapeout/caravel-breakout-pcb

However, since we are going to tape out a USB (CDC ACM) Core on a physical chip, the timeline is pretty strict:

1. Taping out (submitting the final design files to the fab) in September.
2. Chips will be back in February 2024 or so. At that point, we'll be able to test them and finalize the PCB.

Since the VID / PID are hard coded during the chip manufacturing, we need to know them in advance.

tannewt (approved):
Open preliminary versions are fine. Good luck with the tape out!

urish:
Thanks!

PR #868, opened 2023-07-23
PID Request for UniversalboxARM DFU (0x7702)

Solartraveler:


tannewt (changes requested):
Where is the DFU source code? Hardware files look good.

Solartraveler:
Here starts the DFU implementation :)
https://github.com/Solartraveler/UniversalboxArm/blob/main/src/stm32l452/03-loader/loader.c#L206

tannewt (approved):

PR #869, opened 2023-07-29
Improve options for RV003USB

cnlohr:
It would be really helpful if there were 3 basic devices - a bootloader, a HID descriptor and a generic non-HID descriptor for users.

See more about rv003usb here (also referenced in the changed files): https://github.com/cnlohr/rv003usb

* Update the Bootloader description.
* Add generic HID for rv003usb project.
* Add generic USB device for rv003usb project.

### DIFF
--- a/1209/4269/index.md
+++ b/1209/4269/index.md
@@ -1,0 +1,9 @@
++---
++layout: pid
++title: MAGFest 2024 Swadge
++owner: cnlohr
++license: MIT-x11, Public Domain
++site: https://github.com/AEFeinstein/Swadge-IDF-5.0
++---
++The MAGFest Swadge is a hand-held game system / multi-purpose device with an LCD, IMU, buzzers, microphone, touch pad and an ESP32-S2.  This PID is used for debug printf, and sandbox operations.
+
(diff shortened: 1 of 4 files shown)


tannewt (changes requested):
Why rely on serial to differentiate projects?

It seems to me that you really want to use the test PIDs for example code. Actual products should use their own PIDs.

cnlohr:
I would not want a separate PID for each example within my project.  There may be 10 or 15 before I'm done with this!

In general, I think it's fine, because it doesn't matter if there are multiple of the same VID/PID's that are identical when it comes to HID dongles plugged into the same system.

And it doesn't matter if there are multiple of the same for devices that are being used by bespoke code through HIDAPI.

Would it really be preferential for me to make another 5 PIDs right now?

tannewt:
Why not use test PIDs for example code?

cnlohr:
@tannewt  I don't understand the question.  Please elaborate, test codes meaning what?

tannewt:
> @tannewt I don't understand the question. Please elaborate, test codes meaning what?

https://pid.codes/1209/0001/ 0001 - 0010 are private use test codes.

tannewt (commented):
Looks like I've allowed the CH55 folks to have generic IDs for different descriptor combos. So, I think these are ok. However, I wouldn't suggest differentiating devices over serial. Different fixed devices should use different PIDs, not a custom user land implementation. So, please remove that portion of the comments.

cnlohr:
Bleh sorry I missed this for so long.

PR #870, opened 2023-08-05
PID for Serial WS2812 device, ORG for hrmny.sh

ForsakenHarmony:
Working on a PCB design for it, which will be open hardware.


tannewt (approved):
Ok to use this on the Pico since it isn't your board. Please request another PID when you have an open board and don't use this on any closed boards you make.

PR #875, opened 2023-08-12
Requesting PID for RetroZord adapter device

sonik-br:


tannewt (commented):
Are these using arduinos wired to controllers or custom pcbs?

sonik-br:
Hi,
Common retail arduino.
Gamepad direct wired or using an extension cable (to be able to select and use between a number of supported controllers).
Simple wiring instructions are available, like [this](https://github.com/sonik-br/RetroZordAdapter/tree/main/Saturn#arduino-leonardo-pin-mapping).

tannewt (approved):

PR #886, opened 2023-09-21
Requesting PIDs for Solo1 fork

sandeepansg:
Requesting PIDs for a fork of Solokey Solo 1


tannewt (changes requested):
Your repo doesn't actually have any changes from upstream except changing the PID. Please use a test pid until you actually have a functional difference.

PR #888, opened 2023-10-07
Request for PID DDAA for org EllisGL

ellisgl:
Requesting the PID of DDAA for a device that speaks the DDAA protocol.


tannewt (approved):
BSD-3 arduino on third party closed hardware.

PR #890, opened 2023-10-12
Requesting VID for RAVA RNG v1

gabrielguerrer:


gabrielguerrer:
Will fix the org title and resubmit

PR #894, opened 2023-11-02
Add PID for Tangara music player

jqln-0:
Hey there! I hope I did everything right :)

We're making an open hardware media player! The sources are up and it's pretty functional, so this seemed a good time to start using a real USB ID.

The site right now just links to the same project page on sourcehut as the sources link. I'll likely send you another PR in a few months to point it at better documentation, but we haven't written proper docs yet.

We've got a crowdfund pre-launch page up atm, but I didn't link it in any of the files since it felt a little rude to put what's essentially an ad in your lovely repository! It's here if you'd like more info tho: https://www.crowdsupply.com/cool-tech-zone/tangara

Thanks so much for your time, and for setting up this handy resource!

### DIFF
--- a/1209/2014/index.md
+++ b/1209/2014/index.md
@@ -1,0 +1,9 @@
+---
+layout: pid
+title: Tangara
+owner: cooltechzone
+license: GPLv3 and CERN-OHL
+site: https://sr.ht/~jacqueline/tangara/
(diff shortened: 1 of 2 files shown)


tannewt (approved):
Kicad hardware file and cpp source

PR #896, opened 2023-11-06
Add org 'ymkn' and pid '2AF6' for my OSS mechanical keyboard PCB

ymkn:


tannewt (approved):
Kicad HW under MIT. SW is QMK keyboard def.

PR #901, opened 2023-11-19
Requesting PID 0xb420 

dotcypress:
Can I have `0xB420` for https://github.com/dotcypress/neopixel-webusb?


tannewt (approved):
Rust source under Apache or MIT. Third party RP2040 hardware.

PR #902, opened 2023-11-20
Requesting PID for AIRDOS04

roman-dvorak:
Hi, I would like to register a VID/PID for the AIRDOS04 detector. Thank you very much.

### DIFF
--- a/1209/7AA0/index.md
+++ b/1209/7AA0/index.md
@@ -1,7 +1,7 @@
 ---
 layout: pid
 title: AIRDOS04 - Advanced Radiation Monitoring for Aircrafts
-owner: Universal Scientific Technologies
+owner: UniversalScientificTechnologies
 license: GPLv3.0
 site: https://www.ust.cz/
 source: https://github.com/universalScientificTechnologies/AIRDOS04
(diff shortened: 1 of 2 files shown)


tannewt (changes requested):
Where are the open hardware file? I only see 3d printing files and arduino source.

roman-dvorak:
Hi @tannewt, thank you for your comment. Sure, let me know to explain it.. 

The AIRDOS04 detector consists of three electronic boards that together (including printed parts) forms a single unit. We have it designed this way because the data and battery part ([BATDATUNIT01](https://www.mlab.cz/module/BATDATUNIT01/)) is part of the [MLAB](www.mlab.cz{ kit (also GPLv3) and can be used for other devices as well. When used in the AIRDOS04 detector, we would like to have one unique VID/PID. The data and battery board contains an MCU, data storage, llion batteries, and accessories. The second part is the actual detection part equipted with analogue circuits, which is called [USTSIPIN03](https://github.com/kaklik/USTSIPIN03/). 

I have placed these boards in AIRDOS04 repository in form of submodule in folder [/hw/modules/](https://github.com/roman-dvorak/AIRDOS04/tree/AIRDOS04A/hw/modules)


I hope that it is ok and sufficient :-)

tannewt (approved):
Works for me! Kicad source files under GPL3.

PR #904, opened 2023-11-23
Request PID for MAGFest Swadge.

cnlohr:
The MAGFest Swadge is a hand-held game system / multi-purpose device with an LCD, IMU, buzzers, microphone, touch pad and an ESP32-S2.  This PID is used for HID input, debug printf, and sandbox operations.

This is a request to see if we can get a PID code for it instead of re-using the manufacturer's VID/PID pair.

### DIFF
--- a/1209/4269/index.md
+++ b/1209/4269/index.md
@@ -1,0 +1,10 @@
+---
+layout: pid
+title: MAGFest 2024 Swadge
+owner: cnlohr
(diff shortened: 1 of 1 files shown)


tannewt (approved):
Hardware is MIT and Kicad files linked from software repo. Software is ESP-IDF C code.

PR #905, opened 2023-11-30
Add organization (radagast), request pid 1209/AE6E

DavidCPlatt:
Requesting a pid for my OpenTNC project

### DIFF
--- a/1209/AE6E/index.md
+++ b/1209/AE6E/index.md
@@ -4,5 +4,5 @@ title: OpenTNC
 owner: radagast
 license: GPL V3 (firmware) and CC BY-SA (schematic)
 site: http://www.radagast.org/~dplatt/hamradio/OpenTNC
-source: http://www.radagast.org/git/OpenTNC
+source: https://gitlab.com/davidplatt/opentnc
 ---


tannewt (commented):
The source URL doesn't load in my browser. Is there a way to view the repo without git clone?

DavidCPlatt:
It can't be viewed on-line from the repo on my personal server - I don't think the Apache plugin I use for git allows that.

I've been planning to move it over to a more public repository system, and so I've just created a repo at gitlab and pushed it there - that's where it'll live in the longer term, I think.  I was waiting until I got a proper USB ID before doing this... but that's obviously a chicken-and-egg problem, so I've just gone ahead and done it.

You should be able to view or clone the repo at

  https://gitlab.com/davidplatt/opentnc

I'll work at fixing up the README to reflect the new location.

DavidCPlatt:
I've updated the source-repo link in the pid.codes index, and in the README.md file in all of the relevant locations.

tannewt (approved):
Thank you!

C software source. Kicad hardware files.

PR #907, opened 2023-12-06
Request PID 2AF3 - 2AF5 for my keyboard devices

ymkn:


tannewt (approved):
Hardware is KiCad. Software is QMK + other config files in repo. All MIT.

PR #908, opened 2023-12-15
Requesting PID 0xF1DE for Fidelio

danielinux:


tannewt (approved):
GPL Software for anyones RP2040 board.

PR #909, opened 2023-12-15
Add flow3r pid codes F103 and F10B

timonsku:
flow3r is an open source experimental music instrument.
Hardware repository: https://git.flow3r.garden/flow3r/flow3r-hardware
Licensed under CERN OHL S

Software repository: https://git.flow3r.garden/flow3r/flow3r-firmware
Licensed under GPL v3

We need two PIDs to be able to distinguish easily between the device running normally and being in recovery/bootloader mode, which is a graphical second stage bootloader that allows to do a few low level actions and can receive firmware via USB or mount the user data as mass storage.
In normal operation the main use of the USB interface is accessing the Micropython REPL and doing USB Audio related tasks. Software that looks for VID/PID pairs should not try to grab the device while it is in bootloader mode. Having distinct PIDs solves that.

### DIFF
--- a/1209/F103/index.md
+++ b/1209/F103/index.md
@@ -1,0 +1,9 @@
+---
+layout: pid
+title: flow3r-firmware
+owner: flow3r
+license: CERN OHL S and GPL v3
+site: https://flow3r.garden
+source: https://git.flow3r.garden/flow3r/
+---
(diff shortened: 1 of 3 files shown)


tannewt (approved):
Hardware files are Kicad and software is ESP-IDF.

PR #910, opened 2023-12-24
Adding a new org and requesting C000 for a project

hrvach:
https://github.org/hrvach/deskhop is a keyboard/mouse switcher with some internal logic to allow for super-short switch times and mouse cursor tracking, so a mouse cursor can be dragged between different computers and OS-es.

Kindly requesting PID C000 for this project, all source code, schematics, PCB layout and 3D printable case are published under GPLv3.

Thank you!

### DIFF
--- a/1209/C000/index.md
+++ b/1209/C000/index.md
@@ -1,0 +1,9 @@
+---
+layout: pid
+title: DeskHop
+owner: hrvoje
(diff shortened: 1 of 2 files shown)


tannewt (approved):
Hardware files are KiCad. Software is pico-sdk C.

PR #911, opened 2023-12-26
Adding Jumperless to the PID codes list

Architeuthis-Flux:
Added my organization, Architeuthis Flux to the org directory. And chose PID 0xACAB so I added a folder for that with the required info inside.

Thanks for doing this!


tannewt (approved):
Kicad hardware files. PlatformIO C++ MIT software source.

PR #912, opened 2023-12-31
❤️ pid.codes

DxBang:
I have to say this is an awesome idea!

Happy New Year!! 🍾🥂🎆

### DIFF
--- a/1209/B501/index.md
+++ b/1209/B501/index.md
@@ -1,0 +1,8 @@
+---
+layout: pid
+title: Bang Buttons
+owner: BangSystems
+license: MIT
+site: https://bang.systems/
+source: https://github.com/DxBang/BangButtons
+---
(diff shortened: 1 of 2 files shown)


tannewt (changes requested):
Code is PlatformIO C++. I know the microcontrollers are third-party but I see mention of PCBs of your own design. Where are their sources?

tannewt:
> The microcontroller is just an Arduino Leonardo, the custom board is just a daughter board for matrix grid with diodes. I can add a link to it on my website and describe how to build it if required, Since I cannot see a "Share design" on JLCPCB where I had it printed.

Did you design it on the JLCPCB site? Usually there is a way to export it.

The point of pid.codes is to encourage open hardware so I'd like to see the daughter board open hardware.

PR #914, opened 2024-01-07
Request pid (0x8801) for 16F1454 PIC programmer

hanyazou:


tannewt (approved):
Software is MPLAB C. Hardware is Kicad.

PR #919, opened 2024-01-31
ADD request PID:DFAA

marksard:
ADD request PID:DFAA


tannewt (approved):
Software is QMK. PCB is Kicad.

PR #926, opened 2024-02-04
Request PID FF01 for TReK_withTK

digitarhythm:
TReK withTK is Macropad with RP2040-Zero.

KiCAD file and Firmware:
https://github.com/digitarhythm/TReK_withTK


tannewt (approved):
Kicad PCBs. QMK config. Both MIT.

PR #927, opened 2024-02-06
Request PID 4C09 for MISAKANITHM

misaka19465:
MISAKAMITHM comes for touch-type games, with 32 touch keys, 6 sets of hight sensors and 31 RGB LED lights.

The request is only for the firmware, and the board we designed is closed-sourced. However, the VID and PID must match with APIs provided by host device, and it is easy to port to other devices.

### DIFF
--- a/1209/4C09/index.md
+++ b/1209/4C09/index.md
@@ -1,7 +1,7 @@
 ---
 layout: pid
 title: MISAKANITHM
-owner: MisakaNet Team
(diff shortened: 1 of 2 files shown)


tannewt:
> The request is only for the firmware, and the board we designed is closed-sourced.

Sorry, the point of pid.codes is to promote open source hardware and software. We do not provide PIDs for use with closed source, first party hardware because you could open source it.

PR #928, opened 2024-02-16
Request PID 0x2024 for picopico 

zeriyoshi:
`picopico` is a PCB with RP2040 and comes with features to support development with RP2040. This is a development kit for my future projects.

The `picopico` HW design in KiCad is available on [GitHub](https://github.com/zeriyoshi/picopico/). I will also use [QMK Firmware](https://github.com/qmk/qmk_firmware/) as firmware and apply to get PID ~~`0x1994`~~ `0x2024` to use the [VIA Configurator](https://www.caniusevia.com/).


tannewt (changes requested):
Hardware KiCad files under MIT are good.

Do you have a QMK board definition? Please link to it from your picopico repo.

PR #930, opened 2024-02-20
Requesting PID for MMHD

mmkeule:
Hello all

this is my first PID request, i hope i do it right.

thank you in advanced


tannewt (changes requested):
Please include modifiable ecad files under an open license. Please also include what software will be running on the hardware and using the PID.

PR #931, opened 2024-02-25
Request PID 0xFF02 for TReK G50 TIJUANA

digitarhythm:
TReK G50 TIJUANA is 40% keyboard with Tenkey.

KiCAD file and Firmware:
https://github.com/digitarhythm/tijuana


tannewt (approved):
kicad files under MIT. Software is QMK and is linked from HW readme.

PR #932, opened 2024-02-27
PID request (0xFA55) for TurtlPass Firmware 

ryanamaral:


tannewt (approved):
GPLv3 CPP source on third party RP2040 hardware.

PR #936, opened 2024-03-10
PID Code request for Ergodonk split keyboard

JellyTitan:
PID request is for an MIT licensed open-source split ergonomic keyboard.


tannewt (approved):
MIT QMK and kicad files.

PR #937, opened 2024-03-17
Add MelanieT org and PiLomar driver

MelanieT:


tannewt (approved):
MIT pico-sdk source + Eagle files.

PR #942, opened 2024-04-05
Request PID 4C09 for MISAKANITHM

misaka19465:
MISAKAMITHM comes for touch-type games, with 32 touch keys, 6 sets of hight sensors and 31 RGB LED lights.

As #927 was not accepted, I published the hardware as well.


tannewt (changes requested):
GPL3 C++ source looks ok. PDF's are not modifiable and therefore, not considered open enough. Please export the EasyEDA file and include it in the repo: https://docs.easyeda.com/en/Export/Export-EasyEDA-Source-File/index.html

PR #943, opened 2024-04-06
Update URLs and name for `PassPico` to `passta`

brxken128:
This PR adds my personal site ([brxken.dev](https://brxken.dev/)) and updates the URL of the repository, as well as the name of the project as it has since been updated (no boards in production are using the name `PassPico`).

I've decided to move away from the Pi Pico (hence the prior name `PassPico`) and focus on alternative chips and boards that offer higher levels of security, but I still plan to have support for the Pi Pico/RP2040 (just at the user's own risk, unless they add a HSM/cryptographic co-processor, which I'll fully support).

### DIFF
--- a/1209/E128/index.md
+++ b/1209/E128/index.md
@@ -1,8 +1,8 @@
 ---
 layout: pid
-title: PassPico
+title: passta
 owner: brxken128
 license: BSD 2-Clause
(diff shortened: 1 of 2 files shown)


tannewt (approved):
Ok

PR #944, opened 2024-04-19
Request PID FF03 for 'TReK 64 AVAIN'.

digitarhythm:
We have produced a keyboard "TReK 64 AVAIN", please register it with PID "FF03".


tannewt (approved):
QMK and Kicad files look good.

PR #950, opened 2024-05-21
Org and PID Request for Wee Noise Makers PGB-1

Fabien-Chouteau:


tannewt (changes requested):
Please like to the hardware source files from the firmware repo that is linked from here.

Fabien-Chouteau:
Hello @tannewt, I plan to release the schematics of the production boards but it's not done yet.

I got confused by the instruction and I though either open-source software or hardware was enough to qualify.

tannewt:
Nope, the how-to page says:

> If your project involves both hardware and software, both need to be licensed under recognised OSS and OSHW licenses.

I do allow PIDs for open software on third-party closed hardware because the hardware piece is out of control of the software dev. In your case, you are making the hardware and can fully open it for a PID. If you don't want to do that, then see the FAQ for other ways to get a USB PID.

PR #958, opened 2024-07-11
Request PID FF04/FF05

digitarhythm:
Request PID '1209/FF04' for keyboard 'TReK Lagoon' and PID '1209/FF05' for keyboard 'TReK Lagoon SP'.


tannewt (approved):
MIT KiCad and QMK source.

PR #961, opened 2024-08-04
Add UIAPduino Pro Micro CH32V003

YuukiUmeta-UIAP:
That is hardware implementation inspired by cnlohr/rv003usb

### DIFF
--- a/1209/B803/index.md
+++ b/1209/B803/index.md
@@ -1,0 +1,9 @@
+---
+layout: pid
+title: UIAPduino Pro Micro CH32V003
+owner: UIAP
+license: MIT-x11, CC BY-SA 4.0, Public Domain
+site: https://www.uiap.jp/en/uiapduino/pro-micro/ch32v003/
+source: https://github.com/YuukiUmeta-UIAP/rv003usb
+---
+Forked from cnlohr/rv003usb, and Hardware implementation for it.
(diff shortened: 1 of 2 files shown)


YuukiUmeta-UIAP:
Sorry, the directory name was in the wrong case letter.

I have been fixed.

tannewt (changes requested):
Where are the hardware files provided with a license? I see a tar download but it only has the eagle files. Is there a hardware repo with a license?

YuukiUmeta-UIAP:
> Where are the hardware files provided with a license? I see a tar download but it only has the eagle files. Is there a hardware repo with a license?

That is the eagle files. Those files are include references of license URL. Aren,t enough?

tannewt:
> That is the eagle files. Those files are include references of license URL. Aren,t enough?

I don't have Eagle so it's better to have it outside the file. (Usually folks have a git repo with a license file separate.)

YuukiUmeta-UIAP:
> > That is the eagle files. Those files are include references of license URL. Aren,t enough?
> 
> I don't have Eagle so it's better to have it outside the file. (Usually folks have a git repo with a license file separate.)

Thank you for your suggestion. I moved it to github repository and added license file.

tannewt (approved):
Thank you!

PR #972, opened 2024-09-10
Create mnt-research org and request PIDs

mntmn:
Hi,

many thanks for providing this service. So far we've been squatting USB IDs for our open hardware devices, but we'd like to clean up our act and also integrate nicely with fwupd. Because there have been some major revisions of keyboards and trackballs across 3 of our OSHW laptop models, the request contains 10 IDs. I hope that's OK.

Regards,
Lukas / minute

### DIFF
--- a/1209/6D00/index.md
+++ b/1209/6D00/index.md
@@ -1,0 +1,9 @@
+---
+layout: pid
+title: MNT Reform Keyboard 2.0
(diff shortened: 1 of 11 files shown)


tannewt (approved):
C source that is MIT or GPL. Hardware files are open licensed Kicad files.

PR #977, opened 2024-09-23
Request PID  3249 for USB MIDI to Sync Converter

YuuichiAkagawa:
Requesting a PID for USB MIDI to Sync Converter, an open hardware/firmware synchronizer for electronic musical instruments.


tannewt (approved):
Arduino C + Kicad files. GPL3

YuuichiAkagawa:
Thanks

PR #978, opened 2024-09-25
Request PID FAD1 for 16nx Faderbank.

infovore:
Added org page for a holding org, and a PID page for the 16nx Faderbank. The firmware and hardware are currently in separate repos, as the firmware may be used for other devices in future; I've included a link to both in the PID pages. (If it is, build process will ensure appropriate PID codes are used).


tannewt (approved):
Kicad files and C (pico-sdk) source.

PR #979, opened 2024-09-25
Added First.AC and MFSerial at PID 2525

zibredlektab:


tannewt (approved):
Kicad and Arduino. GPL-3

PR #981, opened 2024-10-11
Requst PID 1209:CA02 for the CANnectivity USB to CAN adapter DFU firmware

henrikbrixandersen:
The CANnectivity USB to CAN adapter firmware, which uses 1209:CA01, has the capability of being updated via USB Device Firmware Upgrade (DFU) mode.

This requires a separate USB PID in order not to load the regular USB to CAN adapter driver when booted in DFU mode.


tannewt (approved):
Thanks for the review on my PR.

PR #991, opened 2024-10-24
Request PID BA11 to our Arduino-compatible board for scientific balloon experiment

mzks:
Hardware/software information is here (https://github.com/mzks/balsun).


tannewt (changes requested):
Hardware source looks ok. `arduino_examples` directory is empty so the software isn't open source yet. I assume it'd be what uses the PID so please post it.

PR #995, opened 2024-11-26
Request PID 2070 for VG3D printer control board

GhoweVege:
We would like to get a PID for our 3D printer control board. Links to all sources can be found [here](https://github.com/Vegetronix-Inc/VG3D-Source-Index). Let me know if I've missed anything. Thanks!


tannewt (approved):
GPL software and open third-party hardware.

PR #999, opened 2024-12-02
Request PID 1209:8990-8991 for the project KVM-Card-Mini

Jackadminx:
KVM-Card-Mini project is a simple achieve of the KVM Console to USB function.
I would like to request fixed PID to avoid conflicts with other devices.


tannewt (approved):
MIT C source. MIT EasyEDA pro files.

PR #1004, opened 2024-12-30
Request PID 1209:E501 for a custom keybard

kerikun11:
I would appreciate it if you could review this PR.


tannewt (approved):
GPL v2 QMK. CC BY-NC kicad files.

PR #1008, opened 2025-01-11
Request PID 1209:6660 for Maebe.Me - The Button

maebemae:
https://github.com/maebemae/TheButton is a velocity sensitive macro button which sends relevant messages based on the press velocity.

Currently it is based on the STM32C071 but with a few tweaks it is compatible with a wide range of STM32 boards. Hobbyists can build their own on the STM32c071-nucleo board.

Source code and hardware are licensed under GNU GPL v3

### DIFF
--- a/6660/index.md
+++ b/6660/index.md
@@ -1,0 +1,9 @@
+---
+layout: pid
+title: The Button
+owner: Maebe.Me
+license: GNU GPL v3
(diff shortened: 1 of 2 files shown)


tannewt (approved):
GPL3 CubeMX C code and Kicad files.

PR #1010, opened 2025-01-19
4Project VID/PID request

4project-co-il:
Thank you


tannewt (approved):
MIT Kicad files. MIT Arduino C++.

PR #1013, opened 2025-01-28
Request to add wlanpi org with PID 2042

jolla:


1209/2042/index.md
```diff
@@ -0,0 +1,9 @@
+---
+layout: pid
+title: USB Ethernet
+owner: WLAN Pi
```
tannewt, line 4:
```suggestion
owner: wlanpi
```

tannewt (changes requested):
What source code will use this PID? It looks like there is a lot of different software pieces. Does it run on third party hardware (raspberry pis)?

jolla:
This will be used on Raspberry Pi CM4 based solution. We are creating a custom RNDIS USB ethernet gadget using libcomposite which is why we need a PID so Windows will correctly identify the device as an RNDIS device.

tannewt:
What hardware hosts the CM4? Is it open source?

jolla:
WLAN Pi M4 (CM4 variant): Uses a customized carrier board manufactured by MCUZone, the carrier board has not been open sourced.
WLAN Pi R4 (RPi 4 varient): Uses standard off the shelf Raspberry Pi 4 hardware

tannewt (approved):
:+1:

PR #1018, opened 2025-02-08
Pinscape Pico

mjrgh:
Request to assign PID EAEB to Pinscape Pico (sequel to the original KL25Z Pinscape, at PID EAEA).  This project shares its name with the KL25Z version, but it's original software with a new set of application-specific USB interfaces, so it needs a unique ID so that host software doesn't misidentify it as the KL25Z project.


tannewt (approved):
MIT pico-sdk C++ source. MIT Eagle source.

PR #1024, opened 2025-02-28
Reserving PID CCCC for the zeptoforth CDC Console

tabemann:
Reserving PID CCCC for the zeptoforth CDC Console


tannewt (approved):
MIT assembly code for third party hardware

PR #1026, opened 2025-02-28
Request PID FF06

digitarhythm:
Request PID '1209/FF06' for keyboard 'TReK Periodt'.


tannewt (approved):
MIT kicad files and QMK software.

PR #1030, opened 2025-03-14
Request PID F666 for my USB to EPROM+ adapter.

keshbach:
Request PID F666 for my USB to EPROM+ adapter.


tannewt (approved):
C source. PCB123 hardware.

PR #1031, opened 2025-03-14
Request VID: 1209 / PID: DE20

Badbird5907:
Hi, I'm requesting PID `DE20` under VID `1209` for my mechanical keyboard project Deltaboard.

The hardware (KiCAD) and source code (QMK) files can be found here: https://github.com/Badbird5907/deltaboard


tannewt (approved):
Thanks!

PR #1032, opened 2025-03-15
Add 1209:BC01 as kearney.dev:fat8

jkearneyma:
Allocate a VID:PID of 1209:BC01 to the open-source fat8 project: https://github.com/jkearneyma/fat8


jkearneyma:
typo. will resubmit fixed

PR #1033, opened 2025-03-15
Allocate a VID:PID of 1209:BC01 to kearney.dev fat8

jkearneyma:
Allocate a VID:PID of 1209:BC01 to the open-source fat8 project: https://github.com/jkearneyma/fat8


tannewt (approved):
Kicad and assembly source.

PR #1040, opened 2025-04-06
Added pico2-ice at 1209/B1C1

vr2045:
Added pico2-ice at 1209/B1C1.
Thank you for all the work you do!


tannewt (approved):
MIT Kicad and pico-sdk source.

PR #1046, opened 2025-04-19
Request 1209/2044For KBP Macro Pad

YodaDistro:
Requesting a reservation on `1209:2044` for the KBP Macro Pad.


tannewt (approved):
QMK. Third party hardware.

PR #1047, opened 2025-04-22
Request `1209:3336` for Franky36 added

avdyushin:
Requesting pid for the Franky36 keyboard


tannewt (approved):
Case SCAD. QMK sw. Third party PCB.

PR #1054, opened 2025-05-09
PID: 2121

snappykey:
Hello, I would need the PID 2121 for my project. I don't have pcb for the dev board that uses atmel 32u4. I published the code files for the software on my gh user. Thank you!


tannewt (changes requested):
QMK repo is empty and I'm not sure you have any hardware. Please use a testing PID until the project is further along.

snappykey:
I added the qmk code with "temp" value in pid.  github.com/snappykey/Snappy-Pad-v1-qmk-firmware/

PR #1077, opened 2025-07-29
Added noForth org & PID

WillemOuwerkerk:
Retry for PID B126 without messing with other folders


WillemOuwerkerk:
I tried to add the noForth CDC driver with its own PID B126

tannewt (approved):

PR #1078, opened 2025-07-29
add skji, skji/coquette (1209/FE80)

ember-ana:
Hiya!
This is a PID request for coquette, a keyboard @galaxyeden and I have been working on for the past 7 months (and have been daily-driving for about half of that time!)

It's a labor of love with many incremental improvements and learnings we along the way, that's finally at a point we consider release-ready.

The keyboard is split into board and shield, and we have multiple versions of each based on design preferences, hand measurements, etc.
All boards and shields under the [coquette org](https://thei.rs/coquette) were made from scratch in (mostly) KiCad.

All original creations are licensed under EUPL-1.2, and derivative works under their parent license (MIT, Apache2).

So this doesn't become a novel, the remaining facts & features as bullet points:
- hours of anatomy (human, bird) & ergonomics research
- different shield angles based on hand measurements
- BLE using nRF5340 (instead of the way more common rRF52840 – gets us some better bluetooth perf)
- ridiculous power optimization (2+ months of daily use, depending on battery size)
- LED underglow on board & shield (literally glows because we're using UV LEDs against neon acrylic)
- very pretty PCBs

### DIFF
--- a/1209/FE80/index.md
+++ b/1209/FE80/index.md
@@ -1,0 +1,9 @@
+---
+layout: pid
+title: coquette
+owner: skji
+license: EUPL-1.2
+site: https://thei.rs/coquette
+source: https://thei.rs/coquette
+---
+lovingly over-engineered keyboard (column-stagger, mono-block, split, 42-key) designed to run ZMK with inspiration from: [hummingbird](https://github.com/PJE66/hummingbird), [ruby-choccy](https://thei.rs/rchoc)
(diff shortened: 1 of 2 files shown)


galaxyeden:
I've been the hardware design person on this and this is currently at the state of an open source hardware project anyone can build though there is some firmware work required for some of the features (I'm working on this and plan to have at least mostly feature-complete firmware soon), and because it's now at a state anyone can build it, and we have working hardware, I wanted us to get a real PID code to use, to explain further why! Thanks!

PR #1079, opened 2025-07-31
1209/C00D Modern Cadet keyboard

Cauren:
Oh, no, I commited hardware again!

New device by same maker as 1209/C00C


tannewt (approved):
EasyEDA + QMK

PR #1084, opened 2025-08-16
Request PID F800 for MachiKania Gamepad

kmorimatsu:
I would like to request a PID (0xF800) to use for the gamepad connected by USB to tiny computer system, MachiKania (https://github.com/machikania/phyllosoma).


tannewt (changes requested):
Please fix the CI. Do you need to add an org too?

PR #1098, opened 2025-09-28
heathen.pro - MIDAL MIDI Interface PID request - Kawai GFP-3 compatible pedals standalone MIDI interface

vheathen:
heathen.pro - MIDAL MIDI Interface PID request - Kawai GFP-3 compatible pedals standalone MIDI interface

### DIFF
--- a/1209/4D01/index.md
+++ b/1209/4D01/index.md
@@ -1,0 +1,11 @@
+---
+layout: pid
+title: MIDAL MIDI Interface
+owner: heathen.pro
+license: Apache-2.0
+site: https://github.com/vheathen/midal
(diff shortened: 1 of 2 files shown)


vheathen:
Hi @tannewt , thanks for the review! 

Though the source of truth for the hardware part is the online representation, and this is exactly the reason why I didn't put a snapshot to the repo (it would be much harder to deal with questions regarding hardware when multiple "active" versions available), I put an export actual for a specific date into the repo as you requested and mentioned it the [hardware part](https://github.com/vheathen/midal/blob/master/hardware/README.md#project-introduction).

tannewt (approved):
Thank you!

PR #1100, opened 2025-10-03
Add Sandboxpedal PIDs

DisasterAreaDesigns:
Adding two PIDs for the different PCBs in our open-source DSP guitar pedal projects.


tannewt:
I'll grant you the PID for the software on third party. The other will need to use a testing PID until the hardware is open source.

PR #1103, opened 2025-10-06
added alukovenko org for Dendy compatible USB HID controller

alukovenko:
Bluepill-based device that converts NES Famiclone type controller with serial DB-9 interface into a USB HID


tannewt (approved):
MIT opencm3 code. Third party hardware.

PR #1108, opened 2025-10-22
Kipp Cannon / ALLPRO88 USB

cannon-zz:
- Add an org entry for Kipp Cannon (me)
- Add a PID for an ALLPRO88 USB interface device

### DIFF
--- a/1209/AA88/index.md
+++ b/1209/AA88/index.md
@@ -1,0 +1,9 @@
+---
+layout: pid
+title: ALLPRO88 USB
+owner: KippCannon
+license: GPLv3
+site: https://github.com/cannon-zz/allpro88-fx2-hard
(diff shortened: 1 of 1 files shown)


tannewt (changes requested):
Where is the software source? I see Kicad hardware source files only.

cannon-zz:
> Where is the software source? I see Kicad hardware source files only.

Hi.  The software is at https://github.com/cannon-zz/allpro88-fx2-soft .  That link is near the top of the README.md file in the hardware project.

cannon-zz:
ping.  github makes it seem like changes have been requested, but no change requests have been communicated to me.  if I've missed something, I'm sorry for that.  if you would like something changed, please let me know what, I'm happy to do it.

PR #1111, opened 2025-10-28
register 1209/53A0 for the "Diskomator" by Lukas Rieger

0x53A:
At the moment this is firmware only with off-the-shelf hardware:

* a ESP32-S3 devboard
* a I2S microphone
* a 16x16 Neopixel Matrix, I use one from amazon

This just runs a FFT on the audio and then maps the FFT channels to different patterns.

Configuration can be changed through Bluetooth, and it can act as a USB "Speaker" so it doesn't use the I2S microphone but instead takes audio data through USB. I'd like to also implement USB Microphone, where it pushes the I2S mic recording to  the Host, and firmware updates.

### DIFF
--- a/1209/53A0/index.md
+++ b/1209/53A0/index.md
@@ -1,0 +1,11 @@
+---
+layout: pid
+title: Diskomator 9000 Pro Max
+owner: "0x53A"
+license: MIT
+site: https://0x53a.github.io/esp32-partylight/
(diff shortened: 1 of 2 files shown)


tannewt (changes requested):
Please add a license file to the source repo.

Source is Rust and run on third-party hardware.

tannewt:
Please fix the CI as well.

PR #1166, opened 2026-02-10
Request PID D187 for PewPew uGame-S3

deshipu:


tannewt (approved):
CC BY-SA EasyEDA. CircuitPython (MIT) board def.

PR #1167, opened 2026-02-11
Add CBUnmanned H753 SOM

CBUnmanned:
Allocate PID 5750


tannewt (changes requested):
Where is your hardware source? It needs to be open as well.

PR #1178, opened 2026-03-04
added PID 0x200B (VID 0x1209) for hermes by NyxKeys 

Shiva1796:
Registering NyxKeys, a Canadian keyboard design studio, and claiming PID 0x200B (VID 0x1209) for the Hermes keyboard, a QMK-based custom keyboard.

### DIFF
--- a/1209/200B/index.md
+++ b/1209/200B/index.md
@@ -1,0 +1,10 @@
+---
+layout: pid
+title: Hermes
+owner: nyxkeys
+license: GPL-2.0
+site: https://www.nyxkeys.com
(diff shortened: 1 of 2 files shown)


tannewt (changes requested):
Where is the hardware source?

cnlohr:
To be more specific, the issue is for all pid codes, where there is an existing hardware product, it must be referenced, if there is none then the board files (not just gerbers) must be available somewhere. 

(If I were to take action here, I would just close this issue because there wasn't any update)

❌

PR #1179, opened 2026-03-10
1209/BBD1, an ESP32-S3 BLE-HID serial bridge

clackups:


tannewt (approved):
GPL IDF source. Third party hardware.

PR #1181, opened 2026-03-13
add VID 1209 PID FDFD and new org FutilityDesigns

FutilityDesigns:
Adding new org FutilityDesigns and PID FDFD for Continuum keyboard firmware
RP2040 based keyboard firmware for macropads and custom keyboards. The firmware supports plain key assignments, multi-key macros, layer switching, and per-key OLED displays that show what each key does at a glance. Paired with the Active Window Monitor running on your PC, the keyboard can automatically switch its entire layout as you jump between applications — making it a powerful productivity tool, especially for macro pads.

Hardware documentation in progress, however firmware is designs to run on various hardware and is not locked to one design. Hardware repos will be linked when they're finalized. 

PID is requested for firmware as a way to identify connected devices from the computer side applications.

### DIFF
--- a/1209/FDFD/index.md
+++ b/1209/FDFD/index.md
@@ -1,0 +1,15 @@
+---
+layout: pid
+title: Continuum Keyboard Firmware
+owner: FutilityDesigns
+license: Firmware - MIT, Hardware - CERN-OHL-W-2.0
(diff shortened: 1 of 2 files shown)


tannewt (approved):
MIT platformio. Third party hardware

PR #1185, opened 2026-03-18
Request PIDs f540 and f542 for One ROM

piersfinlayson:
Project page: https://onerom.org

Pre-requisites:

Publicly available source code repository:
- https://github.com/piersfinlayson/one-rom

Containing modifiable PCB design files: 
- kicad files (and other fab files) for all variants
- example: https://github.com/piersfinlayson/one-rom/tree/main/hardware/pcb/verified/fire-24-e/kicad

Source code for a device with a USB interface:
- Main firmware: https://github.com/piersfinlayson/one-rom/tree/main/sdrr
- USB stack: https://github.com/piersfinlayson/one-rom/tree/main/plugins/system/usb

Licensed under a recognized open source or open source hardware license:
- MIT & CC BY-SA 4.0
- License file: https://github.com/piersfinlayson/one-rom/blob/main/LICENSE.md

Edited: 30 Mar 26, PCB files in repo have moved - updated link.

### DIFF
--- a/1209/F540/index.md
+++ b/1209/F540/index.md
@@ -1,0 +1,13 @@
+---
+layout: pid
+title: One ROM Bootloader
+owner: piers.rocks
+license: MIT and CC SA-BY 4.0
+site: https://onerom.org/
+source: https://github.com/piersfinlayson/one-rom/
+---
+Bootloader for One ROM, the most flexible replacement ROM for your retro system.
+
+One ROM emulates a wide range of mask programmed ROMs, EPROMs and EEPROMs, including 24, 28 and 32 pin 8-bit ROMs, and 40 pin 16-bit ROMs.  It can be used to replace the original ROMs used in thousands of types of microprocessor based systems from the late 1970s, 1980s and 1990s.
+
+One ROM is based around Raspberry Pi's RP2350 microcontroller and uses custom, bare metal firmware.
--- a/1209/F542/index.md
+++ b/1209/F542/index.md
@@ -1,0 +1,13 @@
+---
+layout: pid
+title: One ROM
+owner: piers.rocks
+license: MIT and CC SA-BY 4.0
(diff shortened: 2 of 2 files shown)


1209/F540/index.md
```diff
@@ -0,0 +1,13 @@
+---
+layout: pid
+title: One ROM Bootloader
```
tannewt, line 3: Why do you need a bootloader PID? The RP2350 has a built in bootloader.

tannewt (changes requested):
C firmware source and kicad hardware sources look good. One question first.

piersfinlayson:
Thanks for reviewing.

What I would like to do is write a custom bootloader VID/PID into the RP2350's OTP memory on first firmware programming.  This will help both the user and the One ROM tooling to definitively identify the device as a One ROM when in RP2350 bootloader mode - which is most commonly used for "unbricking" the device.

cnlohr:
From what I understand the maintainers don't want to have pid's doubled up for this sort of thing. 

I would let @tannewt make the final call on this one.  I don't know if I would accept this one.

The reason I feel a little weird is, I can sympathize with @piersfinlayson on this one, as there are lots of situations I would prefer to be able to do this.  The compromise I could envision is you would use one bootloader VID/PID for all products you make?  (Would really like more feedback on this one)

❓

piersfinlayson:
Thanks for the comment @cnlohr 

I went with this approach as there appeared to be other RP2040/2350-based projects that have gone with the dual PID approach, one for the bootloader, one for the application.

However, I'm happy to modify the PR for a single (non-bootloader) VID/PID if that's the requirement.

I would be also happy with a more generic VID/PID for all of my open source products' bootloaders using an RP2040/2350.  Again - just let me know if that's what you'd like me to do and I'll modify the PR.

PR #1189, opened 2026-03-24
PID 0x4D58 under VID 0x1209 for Mixlar Mix.

MixlarLabs:
Mixlar Mix is a USB MIDI controller and audio mixer built on ESP32-S3
  using TinyUSB composite device (CDC + MIDI + MSC).

  - Org: MixlarLabs
  - PID: 0x4D58
  - License: MIT


MixlarLabs:
Wrong File

PR #1192, opened 2026-04-02
add scd31 org and rcade display adapter

scd31:
Requesting a PID for my 16kHz arcade CRT display adapter (:


tannewt (approved):
Kicad and Rust source

PR #1193, opened 2026-04-03
Add Woyta org and PID 0x5750 for Woyta-Pad macropad

Woyta64:


tannewt (changes requested):
pico-sdk C source. No hardware source found. Please link to it from the repo that the `source` entry links to.

tannewt (approved):

PR #1194, opened 2026-04-09
PID request for Loopian::QUBIT

hasebems:


cnlohr:
@hasebems ~~you need to include both an organization as well as the PID file.~~  Correction: You do have an org! I just didn't notice.

In addition, you will need to include an actual board design, or pictures of how you have configured the device physically.  I would not be able to look at that site and know how to build or acquire one.  While it's based on a common Seeed XIAO RP2350, it still isn't clear how to use this as a hardware project.

I would leave this open and see if the original author provides feedback.

⚠️

PR #1195, opened 2026-04-11
Request PID FF08

digitarhythm:
Request PID '1209/FF08' for keyboard 'TReK GRAIN'.


tannewt (approved):
QMK source and kicad source.

PR #1198, opened 2026-04-22
Request PID block for Infinite HID controller base (electric, hybrid, diesel, steam)

CDiOS:
Requesting a contiguous block of 4 PIDs for multiple USB HID controller variants based on a shared Arduino Leonardo-compatible platform using the Arduino Joystick Library.

Repository:
https://github.com/CDiOS/infinite-hid-controller-base

Requested assignments:
- 0x5390 Electric Controller Base
- 0x5391 Hybrid Controller Base
- 0x5392 Diesel Controller Base
- 0x5393 Steam Controller Base

These are open reference/base sketches intended as starting points for builders to customize for their own simulation controller hardware.

### DIFF
--- a/1209/5390/index.md
+++ b/1209/5390/index.md
@@ -1,0 +1,9 @@
+---
+layout: pid
+title: Electric Controller Base
+owner: infinite
+license: MIT
+site: https://github.com/CDiOS/infinite-hid-controller-base
+source: https://github.com/CDiOS/infinite-hid-controller-base
+---
+USB HID electric simulation controller base firmware for Arduino Leonardo compatible boards using the Arduino Joystick Library. Provides a simple open reference sketch with one analog axis and example button inputs for further customization.
--- a/1209/5391/index.md
+++ b/1209/5391/index.md
@@ -1,0 +1,9 @@
+---
+layout: pid
+title: Hybrid Controller Base
+owner: infinite
+license: MIT
+site: https://github.com/CDiOS/infinite-hid-controller-base
+source: https://github.com/CDiOS/infinite-hid-controller-base
+---
+USB HID hybrid simulation controller base firmware for Arduino Leonardo compatible boards using the Arduino Joystick Library. Provides a simple open reference sketch with two analog axes and example button inputs for further customization.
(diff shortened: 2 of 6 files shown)


tannewt (changes requested):
I'd rather not grant four PIDs for example code. Example code can use test PIDs. A host can also inspect the HID report descriptor to know how many of each thing is present.

CDiOS:
Thanks for the review, I appreciate the feedback.

To clarify, the repository contains simplified reference sketches for accessibility, but the PIDs are intended for real hardware deployments rather than example-only use. The intent is to group devices into stable top-level controller classes (primarily electric and diesel), with individual models and configurations differentiated through USB descriptors and report structure rather than additional PIDs.

For example:
- 1209:5390 → Electric controller class
- 1209:5392 → Diesel controller class

Specific models within each class would then be identified through descriptors rather than separate PIDs. The example code is included so that users who encounter these PIDs have a working open reference, but it is not the primary purpose of the identifiers. I’m happy to reduce the request accordingly and request two PIDs to represent these core controller classes.

tannewt:
Are the real hardware deployments a specific PCB design?

CDiOS:
The deployments are based on Arduino Leonardo–compatible (ATmega32U4) hardware, not a single fixed PCB design.

The reference sketches reflect a baseline Leonardo-style implementation so they remain accessible and reproducible, but the intended devices are defined configurations of that platform with different control layouts, I/O mappings, and HID report structures depending on the controller class (e.g., electric vs diesel).

So while they share a common USB-capable MCU platform, they are not simply the same device with minor firmware tweaks — each class represents a distinct HID configuration and interface layout.

The goal is to maintain stable identification at the controller class level (via PID), while allowing specific models and variants within that class to be differentiated through descriptors and report structure rather than additional PIDs.

cnlohr:
Pid Codes need to be for hardware embodiments (where applicable).  So, if you were to make some reference boards and publish them or show very clear hardware embodiment's using a dev board, I think we could accept it.  But from my understanding just pointing people at a dev board and example code is insufficient.

For this reason I would close the issue and allow the user to reopen it when they add additional hardware information and/or publish their boards. 

❌

PR #1203, opened 2026-05-02
Added LenweSaralonde organization and BA55 Product ID for Musician MIDI Pico Translator

LenweSaralonde:


cnlohr:
After looking through the links in this PR, it looks like this project qualifies for PID Codes.  While it doesn't have a regular PCB design, it gives very detailed schematics and pictures for constructing this device. Given the opportunity, I would accept this PR.

✅️

PR #1205, opened 2026-05-06
D9D0: Finomnis: Busybeacon

Finomnis:
Requires a custom VID because it is controllable via USB-HID. So we need a PID:VID to distinguish it from other devices.

Software: https://github.com/Finomnis/busybeacon
Hardware: https://oshwlab.com/finomnis/busylight

### DIFF
--- a/1209/D9D0/index.md
+++ b/1209/D9D0/index.md
@@ -1,0 +1,12 @@
+---
+layout: pid
+title: Busylight
+owner: Finomnis
+license: MIT OR Apache-2.0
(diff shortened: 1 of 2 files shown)


Finomnis:
EDIT: Hardware + Software is open sourced now :)

Finomnis:
Blocked until rename; name clashes with https://busylight.com

EDIT: Renamed to BusyBeacon

cnlohr:
I believe this project qualifies for a PID Code.  I reviewed the links and the files and would approve this given the chance. 

✅️

PR #1206, opened 2026-05-11
Neubras - OpenReader Config Standard (ORCS)

neubras-br:
Neubras - OpenReader Config Standard (ORCS) aims to standardize the configuration of USB card readers. The VID/PID is necessary for the software to identify the device in compatible configuration mode, similar to a bootloader. We will soon add C code examples for STM32FXX and ESP32-S3. We also intend to include basic board layouts so that beginner developers can test reading common cards, such as NFC.
Thanks

### DIFF
--- a/1209/CADA/index.md
+++ b/1209/CADA/index.md
@@ -1,0 +1,13 @@
+---
+layout: pid
+title: OpenReader Config Standard (ORCS)
+owner: Neubras
+license: ORCS Attribution License 1.0
+site: http://neubras.com/open-projects/orcs
+source: https://github.com/neubras-br/OpenReader-Config-Standard-ORCS
+---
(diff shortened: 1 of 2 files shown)


neubras-br:
The goal is to create an open standard so that readers in configuration mode (with a specified VID PID) can be configured to the same standard. Open for any manufacturer or hobbyist to implement.

cnlohr:
I don't believe this project qualifies for a PID Codes code.

You would need to have a physical embodiment with reference open source or commercial board / hardware design to qualify.

At least that's my understanding.

I am gonna leave this as a question for the other maintainers to get their opinion.

EDIT: Also for some reason CI got borked on this so they would need to touch the PR to trigger a re-check.

❓

PR #1209, opened 2026-05-14
Init 1209/7C88 as CATS mrrp

Redrield:


Redrield:
Changed the PID to match the protocol-specific hardware id that this device has gotten, for simplicity.

cnlohr:
@Redrield can you include a PDF schematic, or picture of the hardware embodiment.

It's also to make too strong of a heads or tails of it because the main README is only one line long.  Could you elaborate on that?

While I think this project is probably eligible for a PID code, I would like the other maintainers to review to get their opinion. 

❓

Redrield:
@cnlohr Just elaborated on the specifics in the readme file for the project, and pushed schematic pdfs to the repo under Hardware/{v1,v2}/schematic.pdf. One thing I would like to add (didn't feel right putting this in the readme as it's subject to change), right now the only interface between the CATS mrrp and a host device is a flatbuffers-based binary protocol passed over USB bulk endpoints (In future I have plans to expose a BLE interface as well, but the USB interface will still exist). 

Given that, driver applications on the host side need a unique PID to identify devices accepting that protocol.

As an example, the Android-based driver application [currently has a device filter configured to look for 0x1209/0x0001](https://codeberg.org/Redrield/CATSdroid/src/branch/main/app/src/main/res/xml/device_filter.xml), since that's the PID i've been testing with while this PR is pending.

PR #1212, opened 2026-05-18
Request PID A5B3

badjeff:
Request for the PID code A5B3 for an open sourced nRF52840 controller board.

### DIFF
--- a/1209/A5B3/index.md
+++ b/1209/A5B3/index.md
@@ -1,0 +1,9 @@
+---
+layout: pid
+title: Chloroplast Bootloader
+owner: badjeff
+license: CERN-OHL-P v2
+site: https://github.com/badjeff/chloroplast-pcb
+source: https://github.com/badjeff/chloroplast-pcb
+---
+An open source Pro-micro form factor nRF52840 controller board.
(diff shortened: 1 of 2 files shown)


cnlohr:
@badjeff I don't see any source code at the link provided within the PR.  Can you please clarify/update with source?

(As a note-to-myself: The hardware files are fine and look good)

(I would just wait to see if the author updates with additional information and would not accept until they do)

badjeff:
@cnlohr I added additional source of boot loader and firmware to the PR.

cnlohr:
@badjeff I recommend linking between the repos so users can identify what hardware/software is for use with which. 

I just now noticed the link at the bottom of the chloroplast-pcb page, but am not familiar with the way those things connected.

You may want to link https://github.com/badjeff/chloroplast-zmk-module to https://github.com/badjeff/chloroplast-pcb via a readme.

Additionally, chloroplast-zmk-module is missing a license.

Once resolved, I will recommend merge.

badjeff:
I have added a README.md for chloroplast-zmk-module. Btw, I think the firmware source is not related to the requested PID. The VID/PID is only used while the board entering DFU mode as it populates itself as mass storage device. That's all done in Adafruit nRF Bootloader.

PR #1218, opened 2026-06-01
Add pid 2641 (volume knob) to oddbit.com org

larsks:


cnlohr:
@larsks can you please include a schematic maybe also with a picture of the encoder hooked up, since this project is intended to be used with a devkit?  Pid Codes tries to have a hardware embodiment for all the PID codes.

PR #1220, opened 2026-06-03
Create index.md for requesting 1209/2348

till-s:


cnlohr:
@till-s it would be great if you could include pictures and PDF versions of schematics for this project.

But, I think it would still qualify for PID codes.  All designs and source is available and open source.

I recommend accepting this project.

✅️

till-s:
Thanks - yes, adding pictures and also PDFs is on my list (all the kicad sources are already up there)

PR #1226, opened 2026-06-16
Add 1209/F1D2 RS-Key

TheMaxMur:
Open-source FIDO2/WebAuthn security-key firmware for the RP2350, written in Rust — also speaks OpenPGP, PIV and OATH.

### DIFF
--- a/1209/F1D2/index.md
+++ b/1209/F1D2/index.md
@@ -1,0 +1,9 @@
+---
+layout: pid
+title: RS-Key
+owner: TheMaxMur
+license: AGPL-3.0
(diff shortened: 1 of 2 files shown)


TheMaxMur:
@tannewt can u check pls?

cnlohr:
This project doesn't fall under the criteria that I got from the maintainers for reviewing projects.  As it's a hardware-agnostic but very specific software project.  Normally PID Codes want to have a hardware embodiment.  I need further direction before I have an opinion on this project. 

I WOULD LEAN towards accepting it. But would not without further direction.

❓

PR #1232, opened 2026-06-26
Allocate PID 0xF1C0 for AMOVLAB Flycore

amovlgf:
This PR requests a pid.codes USB PID allocation for the AMOVLAB Flycore flight controller.

AMOVLAB Flycore is being upstreamed to PX4 in:
https://github.com/PX4/PX4-Autopilot/pull/27407

PX4 reviewers requested that the board stop using Auterion's USB VID `0x3185`. Flycore therefore needs a public and traceable USB VID/PID assignment for its PX4 bootloader and PX4 firmware USB CDC ACM interfaces.

Requested allocation:

- VID: `0x1209`
- PID: `0xF1C0`
- Project: `AMOVLAB Flycore`
- Vendor/organization: `AMOVLAB`
- Usage: PX4 bootloader and PX4 firmware USB CDC ACM device identification for the AMOVLAB Flycore flight controller

Once this PID is allocated, the PX4 Flycore board configuration will be updated to use `VID=0x1209` and `PID=0xF1C0` instead of Auterion's VID.

PR #1233, opened 2026-06-30
Allocate PID 0x2057 for DroneerX6

wszy2021:
This PR requests a pid.codes USB PID allocation for the DroneerX6 flight controller.

DroneerX6 is being upstreamed to PX4 in:
https://github.com/PX4/PX4-Autopilot/pull/27745

PX4 reviewers requested that the board need VID/PID . DroneerX6 therefore needs a public and traceable USB VID/PID assignment for its PX4 bootloader and PX4 firmware USB CDC ACM interfaces.

Requested allocation:

VID: 0x1209
PID: 0x2057
Project: DroneerX6
Vendor/organization: droneer
Usage: PX4 bootloader and PX4 firmware USB CDC ACM device identification for the DroneerX6 flight controller.


cnlohr:
I don't know how I could review this, being a PR instead of a normal project.  It's not immediately evident from that link what hardware this is for, and what the exact software package is.  Is there a better, in-git-three link you could use?   See also #1232

wszy2021:
> I don't know how I could review this, being a PR instead of a normal project. It's not immediately evident from that link what hardware this is for, and what the exact software package is. Is there a better, in-git-three link you could use? See also #1232

how about this
flight controller：https://github.com/Droneer-uav/Droneerx6/blob/main/Droneer%20X6_01.png
Source code: https://github.com/wszy2021/PX4-Autopilot/tree/droneerv1
Licenced under BSD 3-Clause.

PR #1236, opened 2026-07-02
ohmlab 8008 & 8009 PIDs

hamishmorley:


hamishmorley:
hi i'm just wondering when this will be approved?

cnlohr:
@hamishmorley 

1. Can you clarify what the difference is between the FSBL and the application?
2. The link only provides the hardware, and for a single product.  Is there open source software?  If so where is that?
3. You will need to provide hardware sources (i.e. Altium, Eagle, Kicad) for the boards.

Charles

PR #1239, opened 2026-07-05
Add PID 8251 for usbipd-win

dorssel:
This is for the USB/IP stub device of https://github.com/dorssel/usbipd-win.
Driver code is being developed at https://github.com/dorssel/usbipd-win-drivers.


cnlohr:
Software (mix of C# and C) GPL-3.0

There is no hardware with this project, but it makes sense why, being a virtual device for Hyper-V and WSL2.

While I still need clarification if this type of project is OK, if left to my own, I would approve this.

❓

dorssel:
Yes, this is a virtual device. Very similar to https://github.com/pidcodes/pidcodes.github.com/blob/master/1209/8250/index.md, which is for client side, whereas this is for server side. In fact, I chose an adjacent number because they are kind of related.

This driver is supposed to replace the VirtualBox driver. Oracle uses 80EE:CAFE, where 80EE is the Oracle vendor ID and CAFE is probably a reference to Java. However, I cannot use that VID:PID, because the driver interfaces are totally different and the technologies should even be able to coexist.

PR #1242, opened 2026-07-09
Add Left Shift Logical LLC and PIDs 5AFC-5AFF for RecoverKit devices

fcjr:
Requesting 4 PIDs (5AFC–5AFF) for Left Shift Logical LLC's RecoverKit line: USB-PD dongles and an all-in-one appliance that DFU-restore Apple Silicon and T2 Macs.

- Firmware and host stack (Apache-2.0): https://github.com/fcjr/restorekit
- PCB design files (KiCad, under active development): https://github.com/fcjr/restorekit/tree/frank/dongle/hardware/dongle-lite-hw

**Why the Proto Lite (5AFC) and Dongle Lite (5AFD) get separate PIDs:** the Proto Lite is a documented breadboard build (stock Raspberry Pi Pico, FUSB302B breakout, discrete level translators) that a community group assembles themselves from off-the-shelf parts, while the Dongle Lite is the production PCB. The two differ in hardware — the proto has no USB hub path and a different pin map — so host software needs to tell them apart at enumeration.

**The other two:** 5AFE (Dongle Pro) adds a USB-PD charging port; 5AFF (RecoverKit Pro) is an all-in-one appliance with a compute module and four PD-capable USB-C ports that drives four targets in parallel with the same firmware family and vendor interface.


fcjr:
got pid elsewhere

PR #1257, opened 2026-08-09
Add 1209/BBD2, update URLs in 1209/BBD0, 1209/BBD1

clackups:


cnlohr:
For chahor_rotary_keyboard, I think everything's fine. Open source hardware+software, includes PDF, etc.

For esp_hid_serial_bridge, there is no hardware embodiment which is generally requested for pidcode eligibility. 

I would lean towards accepting this PR.  But it will require additional review.

❓

clackups:
esp_hid_serial_bridge is a firmware that will run on any esp32s3 device that has a USB port. This device, for example, is pretty minimalistic, but there are dozens of other options:  https://lilygo.cc/products/t-dongle-s3

As for the new device (BBD2), the repository has the 3D model for the enclosure, and I'll add the schematics and assembly instructions soon.

clackups:
BBD0 and BBD1 were prebviously approved, I just moved the repositories into a new organization, so that they are easier to find.

